PR #645, opened 2015-04-21
add missing CREDITS.txt to KoboRoot.tgz

brunotl:


pventafridda:
something is wrong with this commit..

PR #655, opened 2015-04-29
Nmea sounds4

jaaaaf:
Sounds implementation for external device by sending nmea sentences.
How it works is described in the forum here: http://www.postfrontal.com/forum/topic.asp?TOPIC_ID=7932
Code reviewed by Bruno.
Tested on GoflyV4.


pventafridda:
I think long work to be made easier.
First, we create a GoFly instrument.
Second, when a user selects the gofly instrument it is automatically using also the Sounds.
If he does not want sounds, he disable sounds in configuration.
No need to add any other configuration.

PR #674, opened 2015-05-22
few fix

brunotl:


pventafridda:
I got two crashes, not sure it depends on this commit. I put on hold until I can replicate the problem..Happening on zoom out or zoom in apparently.

PR #753, opened 2015-10-27
last update

brunotl:
- fullscreen on Windows PC.
- kobo can be used as serial device.
- kobo power off if exit from flight or sim mode.
- few other minor fix.


SergioDaSilva82:
Bruno, as I said to you before, I hope you can add to the makeall file the option to create a .exe file with Full Screen mode ON, good option for tablet running Windows 8 or 10: https://www.facebook.com/LK8000/posts/851519591591323

LK8000-PC-FS.exe could be a good idea.

PR #758, opened 2015-11-16
fix freetype TextSize, last char can be clipped with italic font.

brunotl:
Before :
image
After : 
image


pventafridda:
I had not noticed this problem so far, good solution!

brunotl:
3 commit added : major fix for portrait mode.

PR #760, opened 2015-11-24
Radio active clenup

AlphaLima:
This is the Radio Control Interface ported to Linux,
I have added two drivers for the Dittel KRT2:
One is the direct link (non NMEA) driver using a direct binary Dittel protocoll, for this I had to establisch non string bypass function. The KRT2 protocol is working vice versa, So a second instance of LK can act as a Radio hardware emulator, when using a  (virtual) Null modem.

The other driver is an generic NMEA prtocol PVCOM mixed to the NMEA data stream, at the moment I'm working on a NMEA filter module base on an low cost arduino mini. This can easily be interfaced to other properity radio devices.
To mix the PVCOM data, the must be activated an set to the same COM Port as the GPS source.

IMPORTANT: If none of the drivers is activated, it has no effect on the code at all.

On 1.1.2018 we all must have a 8,33kHz radio. 

Here is a little Demo of the function:
https://www.youtube.com/watch?v=GoSxlQvuAhg

Here is the LX approach:
https://www.youtube.com/watch?v=S1UNYIm2_4g&feature=youtu.be&list=PLH8uYt3fuY9GRAKTLwk65lRdLRiwZnUPF


brunotl:
Replaced by #766

PR #762, opened 2015-11-26
switchable optimizer mode regarding this discussion

AlphaLima:
http://www.postfrontal.com/forum/topic.asp?TOPIC_ID=7234
Display2/3 -> Optimizer mode
switches between
"Start on turnpoint"
an 
"Start on Leg"
optimisation
Also added this to custom key and custom Menu respectively
Not only for paragliders very interesting

Here we can see the difference:
The standard "Start on turnpoint" is useless in this case.
turn1

Here we see that "Start on Leg" already finds a FAI triangle in the same situation
leg

### DIFF
--- a/Common/Data/Language/DEFAULT_MENU.TXT
+++ b/Common/Data/Language/DEFAULT_MENU.TXT
@@ -815,10 +815,9 @@ location=7
 mode=Display2
 type=key
 data=9
-event=Null
-label=$(AC03)
+event=Service OPTIMODE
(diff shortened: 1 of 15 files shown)


pventafridda:
If this is affecting also paragliders, we need to know if bruno agree, of course.

pventafridda:
p.s. start on leg was long awaited, thanks!

brunotl:
no objections, look good !

PR #766, opened 2015-11-27
Radio active cleanup

brunotl:
replace Pull-Request #760

_This is the Radio Control Interface ported to Linux,_
_I have added two drivers for the Dittel KRT2:_
_One is the direct link (non NMEA) driver using a direct binary Dittel protocoll, for this I had to establisch non string bypass function. The KRT2 protocol is working vice versa, So a second instance of LK can act as a Radio hardware emulator, when using a (virtual) Null modem._

_The other driver is an generic NMEA prtocol PVCOM mixed to the NMEA data stream, at the moment I'm working on a NMEA filter module base on an low cost arduino mini. This can easily be interfaced to other properity radio devices._
_To mix the PVCOM data, the must be activated an set to the same COM Port as the GPS source._

_IMPORTANT: If none of the drivers is activated, it has no effect on the code at all._

_On 1.1.2018 we all must have a 8,33kHz radio._

_Here is a little Demo of the function:_
_https://www.youtube.com/watch?v=GoSxlQvuAhg_

_Here is the LX approach:_
_https://www.youtube.com/watch?v=S1UNYIm2_4g&feature=youtu.be&list=PLH8uYt3fuY9GRAKTLwk65lRdLRiwZnUPF_


brunotl:
@pventafridda : create a merge commit for clean history,

PR #774, opened 2016-01-08
Fix Freetype TextSize

brunotl:
horiBearingX+Width can be greater than advance, in most case for Oblique font.
this fix need to be done only if advance is lower than horiBearingX+Width, otherwise, we broke right aligned text.

append to commit eb3dfc1b499dca639558dd23bd56a9938c325b87


pventafridda:
There is a problem with calculation of text size now. If you go on MM 4, testing 800x480, some letters are now exceeding the squared boxes.

pventafridda:
I am not sure it was worth the pain to fix all side effects, fixing the italic.. ehm.

brunotl:
not related to text size but related to text clipping, fix in progress ...

pventafridda:
ok thx.

Issue #635, opened 2015-03-29
Kobo: Click on UTM message box doesn't work


SergioDaSilva82:
Paragliding mode with no Task, long click on the compass (top right corner) give us UTM coordinate info.

Before the message box would close if we click on top of it but now it doesn't work.

SergioDaSilva82:
Thank you for the quick fix!

pventafridda:
Maybe you can say "Kobo" only if it does work on Linux but it does not work on kobo.

SergioDaSilva82:
I wrote Kobo to tell where I tested it. Kobo is linux...

brunotl:
all platform, fixed

PR #806, opened 2016-02-08
last fix

brunotl:


SergioDaSilva82:
Don't you think this fix will bring this bug back? https://github.com/LK8000/LK8000/issues/635

AlphaLima:
I can confirm both bugs are fixed without any of the know sideeffects.

AlphaLima:
Sory Sergio, 
you're right tis fix brings back this  #635 UTM bug on a KOBO in Para mode
@brunotl 
maybe we must use the KOBO Compiler directive to separate the two effects.
@pventafridda 
I'm afraid it must be reopened

PR #820, opened 2016-02-24
Why this delay?

AlphaLima:
This delay is anoying in PC Version. Why do we need it. Sound works good without it. Even if it wouldn't do we really need a sound here?

### DIFF
--- a/Common/Source/Draw/MapWindow_Events.cpp
+++ b/Common/Source/Draw/MapWindow_Events.cpp
@@ -113,7 +113,7 @@ start_search:
     UnlockTaskData();
 
     #if (WINDOWSPC>0)
-    if (EnableSoundModes) Poco::Thread::sleep(1000); // let the sound be heard in sequence
+  //  if (EnableSoundModes) Poco::Thread::sleep(1000); // let the sound be heard in sequence
     #endif
     if(dlgGetNoElements() ==0) { 
         if(dyn_range < 120000) {


AlphaLima:
I see no need for the delay

pventafridda:
Tomorrow I check too. It is only on PC, but sound is in any case played after the delay of 1s, so I dont understand why too. Bruno maybe can tell .

brunotl:
See Commit : 804a7095e73ad35a390e523338f16ee3bd274fc4

AlphaLima:
Still don't understand the purpose of the delay in this commit.

pventafridda:
The purpose is to give the same timing you have on pna also on pc. You 
click, you wait while LK is calculating and then you get the answer after a 
beep. But on pc it is instant and does not work as on a pna . So the delay 
will compensate just like with the Oracle.
---
Sent from mobile



Il 25 Febbraio 2016 10:28:18 Ulli Heynen <notifications@github.com> ha scritto:

AlphaLima:
So you want people to get the impression LK is slow even on a GHz PC?

pventafridda:
Ok I tested the differences. Normally on a PNA you long click and get a long click feedback , and immediately after you get the beep confirmation . If you remove the delay, we shall not be able to play the first sound. For me it is not a problem to leave or to remove this delay, I see no difference and no problems in both cases.

AlphaLima:
Well this leads to another problem long click while the map is moving is not a good idea as the objects move under the finger while clicking.
My preference would be long clicks for zooming, and short for select in maps, but that is in conflict with the turbolence proof selection concept.

pventafridda:
It is not a matter of turbulence proof, it is a matter of guessing what is most likely that a user does when touching the screen.
Me, I zoom. And some times during the flight I ask for a multiselection. Most times I dont need it because (no airspace concerns..) all info about nearby waypoint are in multimap page 4, so I might not be a good example.
But in any case, LK since y2009 works that you touch the screen and it zooms.
Doing a revolution on V6 at a few weeks from delivery is crazy but I think it is possible to make it configurable of course.
Although, not a quick and easy think to do, given the complicated way we manage all touches .

AlphaLima:
Right!

pventafridda:
So Bruno what do you think in the end? Should we merge or not??

brunotl:
no good reason to slowdown PC versions by useless Sleep, so yes.

AlphaLima:
same for Oracle?

Am 25.02.2016 um 23:44 schrieb Bruno de Lacheisserie:

pventafridda:
no, let it be like that.. oracle is quite slow on PNA, lets not give false illusions to end users

PR #824, opened 2016-02-25
TCP device port

brunotl:


pventafridda:
Do you want to put this in tomorrow's beta or do you want to keep it for 5.3k ?

brunotl:
Has you want, but soon it's tested soon we will know if something is wrong

pventafridda:
But is it testable? I did not merge it yet so I ask without knowing what it does.
Can we feed LK from a smartphone already? Or how do we test it

brunotl:
I have made test with gpsfeed+ And 2 kobo
All work fine.
Not tested on pna because I don't have device with ip stack.

pventafridda:
Then LETS GO in this beta!

PR #828, opened 2016-02-26
[fix #827] segfault in DrawTerrain if Rect is too small.

### LINKED ISSUE #827: borderline crash
DrawTerrain.cpp 
trenderer exists but terrain_ready is false because TerrainRenderer was aborted while checking for rc.right and rc.bottom compliancy (line 111, around).
We then call CloseTerrainRenderer and we have a crash while doing
delete trenderer

brunotl:


brunotl:
replaced by #829

PR #847, opened 2016-03-15
Fix #837

### LINKED ISSUE #837: Oracle no more working correctly
In V6 the oracle is ignoring nearby items, and changing response for the same position report. By  rotating the map (turning or changing bearing in sim) on the same place, response is changing. It is ignoring sometimes the near airfield, or the near topology item, or another waypoint. 
Changing zoom level will cause the oracle to report far items randomly.
This is happening on all versions. 
Since this is adopted by pilots for precise position report to traffic controllers and for safety, it is a critical problem.

brunotl:


pventafridda:
It is much better, and topology search seems ok now, but we still have a problem with waypoint search.  Nearest airport is not shown on high zoom if it is not close enough. Zooming out will show it again.

PR #848, opened 2016-03-15
Fix

brunotl:


Common/Source/Dialogs/dlgWayQuick.cpp
```diff
@@ -132,7 +133,7 @@ short dlgWayQuickShowModal(void){
   if (WPLSEL.Format == LKW_CUP) {
         TCHAR ttmp[50];
         // and it is landable
-        if ((WPLSEL.Style>1) && (WPLSEL.Style<6) ) {
+        if ((WPLSEL.Style >= STYLE_AIRFIELDGRASS) && (WPLSEL.Style <= STYLE_AIRFIELDSOLID) ) {
```
pventafridda: Why not the same >1 ? Style 1 is not landable. We should respect <= etc. , no?

pventafridda:
Originally we were searching among the farvisibles, now we are searching going through the entire waypoint list. But we are no more inside draw thread, so probably it is not a problem, we can wait (less than before in any case, right?)

brunotl:
right in V5 wait time are 2sec + search time
now it's only search time

Common/Source/Dialogs/dlgWayQuick.cpp
brunotl: no problem, it's more clear like that STYLE_AIRFIELDGRASS are first landable and STYLE_AIRFIELDSOLID are last landable.

PR #849, opened 2016-03-17
Fix

brunotl:


Common/Source/MapDraw/DrawTerrain.cpp
```diff
@@ -533,9 +533,9 @@ class TerrainRenderer {
                             dd1 /= scale;
                             dd2 /= scale;
 
-                            int mag = (dd0 * dd0 + dd1 * dd1 + dd2 * dd2);
+                            int mag = isqrt4(dd0 * dd0 + dd1 * dd1 + dd2 * dd2);
                             if (mag > 0) {
```
pventafridda: Doesnt isqrt4 always return a positive number? In such case we shall never use the rest of the code
} else {
						*imageBuf = oColorBuf[h];
					}
so something is wrong here.

brunotl: in math definitions, sqrt(x) are positive AND negative but it's not the subject.
in computeur world sqrt is always positive, i don't know exception.
in all case isqrt4 return  unsigned int, so for isqrt4(x), if x is positive signed int, result can't be greater than x, then no overflow and result can't be negative in any case.

maybe the explanation are not perfect, but result is same.

PR #877, opened 2016-04-13
new Open Airspace entries AF AG

AlphaLima:
some countries (NL,D, B) publish official open airspace files with the entries
AF: airspace radio frequency
AG: Airspace ground station name
This patch reads these caption data and adds it to the airspace names. 
(Important FlyIn zone switch by AF not affected and still working)

### DIFF
--- a/Common/Source/LKAirspace.cpp
+++ b/Common/Source/LKAirspace.cpp
@@ -28,7 +28,6 @@
 #define DEBUG_AIRSPACE
 #endif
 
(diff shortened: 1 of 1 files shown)


pventafridda:
But... the second patch is undoing the first one. Whats up?

AlphaLima:
second patch, done by mistake!
should I do a separate commit or can you merge the first one only?

Am 13.04.2016 um 23:23 schrieb Paolo Ventafridda:

pventafridda:
i do it

PR #881, opened 2016-04-16
RMZ was read a 'R' (restricted)

AlphaLima:
now changed check RMZ before R


pventafridda:
We cannot merge this change because it is entirely replacing the old file, loosing any track of any change.

PR #915, opened 2016-09-17
New infoboxes

ToninoTarsi:
As discussed in the forum I have added 2 new infoboxes.

1)	[Maximum efficiency speed to fly] 
This is the speed to fly for maximizing distance in the current wind and netto vario conditions.
This is not MacCready STF (neither MC=0) that on the contrary tend to minimize task time and does not depend on current head wind conditions (except that on final glide). 

Maybe a border line information but anyway an important parameter in the STF theory.

2)	[Target Req. Efficicency]
Required efficiency to the current target

This is necessary to fix a problem in the HG/PG mode where there is no way, at the moment, to have this information neither on the overlay nor in the bottom bar.

In my local repo I also have some other  infoboxes suggested by Bruno ( Target Distance .. ) but they are not really necessary  as that information are on the default overlay content. Let me know if you prefer me to commit also this.

Regards
Tony

### DIFF
--- a/Common/Source/Draw/LKProcess.cpp
+++ b/Common/Source/Draw/LKProcess.cpp
@@ -870,6 +870,15 @@ goto_bearing:
 			_stprintf(BufferUnit, TEXT("%s"),(Units::GetHorizontalSpeedName()));
 			_stprintf(BufferTitle, TEXT("%s"), Data_Options[lkindex].Title );
 			break;
+
+		// Speed of maximum efficiency
+		case LK_SPEED_ME:
+			value=SPEEDMODIFY*DerivedDrawInfo.Vme;
+			if (value<0||value>999) value=0; else valid=true;
+			_stprintf(BufferValue, TEXT("%d"),(int)value);
+			_stprintf(BufferUnit, TEXT("%s"),(Units::GetHorizontalSpeedName()));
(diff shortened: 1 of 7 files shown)


Common/Source/Draw/LKProcess.cpp
```diff
@@ -870,6 +870,15 @@ bool MapWindow::LKFormatValue(const short lkindex, const bool lktitle, TCHAR *Bu
 			_stprintf(BufferUnit, TEXT("%s"),(Units::GetHorizontalSpeedName()));
 			_stprintf(BufferTitle, TEXT("%s"), Data_Options[lkindex].Title );
 			break;
+
```
pventafridda: At the beginning I wrote:
Keep this list sorted, so that the compiler can use a jump (branch) table ..
so the best place to place item 133 is after 132, not after 42.. no?

ToninoTarsi: Thanks Paolo. I missed it. Fixed with new commit.

pventafridda:
great thanks. tonight i merge!

Common/Source/Draw/LKProcess.cpp
pventafridda: Done, please remember next time to create a separate branch, in order to avoid pushing your master. It can be bad sometimes.

PR #917, opened 2016-09-21
OpenAIP support

alus-:
Version 6 is now out so I make my pull request for the OpenAIP support.
Differently from what proposed on the forum I put back the "NOGLIDER" definition because it's actually supported by the OpenAir standard so, for now, I think it's better to keep as it is and maybe to add later now airspace definitions at the end.


pventafridda:
Thanks Alberto. Please remember to put down some docs for the users, so they can understand what's new.

PR #919, opened 2016-10-07
fix segfault in ClimbAverageCalcultor with gcc 4.7

brunotl:
regression commit : 0c68bf650354fba119cbab9c7ed0d0f74890f4a5


pventafridda:
Is this going inside 6.0b as well?

brunotl:
no 6.0 is not concerned...

PR #923, opened 2016-10-14
Aircraft symbol

ToninoTarsi:
As discussed here ( http://www.postfrontal.com/forum/topic.asp?TOPIC_ID=8351 ) this permits to change the glider symbol in PG/HG mode choosing from :

- Triangle ( the old and default symbol )
- Paraglider
- Hanglider
- Glider
- Aircraft

It also add some more option in the PgCruiseZoomInit parameter as, the old, (  maximum  1.5km ) was  a bit too low for HG.


Common/Source/Dialogs/dlgConfiguration.cpp
```diff
@@ -1840,7 +1840,7 @@ DataField* dfe = wp->GetDataField();
   if (wp) {
     TCHAR buf1[32];
     DataField* dfe = wp->GetDataField();
-    for (int i=0; i<=9; ++i) {
+    for (int i=0; i<=14; ++i) {
```
pventafridda: Why?

Common/Source/Draw/MapWindowZoom.cpp
```diff
@@ -119,6 +119,11 @@ double MapWindow::Zoom::GetPgCruiseZoomInitValue(int parameter_number) const
     case 7: return 0.75; 
     case 8: return 1.00; 
     case 9: return 1.50;
+    case 10: return 2.00;
+    case 11: return 3.50;
+    case 12: return 5.00;
+    case 13: return 7.50;
+    case 14: return 10.00;
```
pventafridda: Please tony do separate commits for different purposes, in the future.

ToninoTarsi: Ok Paolo., sorry. I will.

PR #935, opened 2016-11-25
Sound table

ToninoTarsi:
Added a default behaviour for the output sound messages sent to the serial port of Kobo.
If SOUND_TABLE.TXT is not found on the _Configuration folder sound messages will be sent as :
$LKALARM,SOUND_CODE_ID
Where SOUND_CODE_ID is the one defined in sound_table.h ( typedef enum sound_code )
If SOUND_TABLE.TXT is found the custom messages in that file  will be used.

### DIFF
--- a/Common/Source/Sound/ExtDev/sound_table.cpp
+++ b/Common/Source/Sound/ExtDev/sound_table.cpp
@@ -31,11 +31,20 @@ bool sound_table::init() {
     TCHAR srcfile[MAX_PATH] = {};
      _stprintf(srcfile,_T("%s" LKD_CONF DIRSEP LKSOUNDTABLE),LKGetLocalPath());
     
-    FILE *fp;
-    if ( (fp=_tfopen(srcfile, _T("rt"))) == NULL ) {
-        StartupStore(_T("...Cannot load conversion sound table file for external sound : %s" ) NEWLINE ,srcfile);
-        return false;
-    }
+     FILE *fp;
+	 if ( (fp=_tfopen(srcfile, _T("rt"))) == NULL ) {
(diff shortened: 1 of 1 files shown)


pventafridda:
OK, I just wonder what is wrong in sending out the sound name instead of a number, in the case we dont have to transcode anything.

ToninoTarsi:
I think nothing is wrong using names.  Codes are  just easier to compare but I do not have any preference. 
If we use names a different implementation would be preferable modifying only ExtDev->Sound.cpp::LKSound(..)
The advantage of this solution will also be that, if we add a sound, there will be no need to add it to sound_table.txt
If you prefer this solution close this one and I will open an other one.

pventafridda:
Yes codes are easier to compare than names. In an external application we receive a number and we handle the number playing a sound.
We can assign the same sound to different numbers, like we do now for piezo speakers.
The point is that this is a classic borderline exception managed as a rule, in other words: made the opposite as it should be done.
Because the rule is: we send out the name or the number of the file to play, and the app will play it.
The exception is: device-app cannot play a file and need to invent something: but that’s a problem of the device-app, not of LK.
What is the point of having LK configure internally what an EXTERNAL device-app should do by its own? None of course.
Either LK or the external device must transcode sounds. And in fact, we have a transcoding table for sounds.

Now, I of course much prefer the correct solution and in fact I had a few months ago implemented it, but never merged.
Why? Because in the end, who is playing sounds externally? Only 1 device for kobo. We can live with that. It is also true that itìs unlikely we are adding sounds to LK...
So I merge your latest commit. No reason to change anything , all considered. I just  wanted to explain my concern.

ToninoTarsi:
Ok

PR #986, opened 2017-01-31
Android

brunotl:


pventafridda:
zoom:RealScale is changed, old 7.5km is now 10km.
Object size is 1.3 times smaller, approx.

PR #988, opened 2017-02-03
lot of fix

brunotl:


pventafridda:
ScreenScale does not change, so it would be nice to have a float screenscale too, ready to use where floats are used. Unless you think that   (x*ScreenScale) >>10  is faster than accessing a float.
Beside, we cannot rescale under 240 pixels now because ScreenScale is forced to be min 1.
Yesterday I had fixed it to go down to 0.84. Is there a reason why you decide that it is wrong?

brunotl:
we can have overload for float, speed will be almost the same if device have fpu, but that avoid rounding, for now it's useless, because we don't use Float for Screen coordinate.

i have removed rescale under 240, because if scale factor is less than 1, IBLSCALE(1) = 0,
so if we do that, we need to check 731 line of code in 74 fles, to huge work for nothing

pventafridda:
you are right. I have spent some time to change it to 0.84 yesterday 
because I do work for nothing and you are also right no need to ask me.


In data 3 Febbraio 2017 10:32:39 Bruno de Lacheisserie 
<notifications@github.com> ha scritto:

pventafridda:
we use float in drawrunway another useless work of mine


In data 3 Febbraio 2017 10:32:39 Bruno de Lacheisserie 
<notifications@github.com> ha scritto:

pventafridda:
speed is almost the same if device has fpu. interesting: for 7 years we 
have been running on 300mhz CPU without problems. now you remove lookups 
tables and avoid to use a simple float global because of speed issues while 
we are moving from 300mhz to 2ghz processors mostly used for gaming!!
we are doing a contest of code beautified maybe that I don't know.




In data 3 Febbraio 2017 10:32:39 Bruno de Lacheisserie 
<notifications@github.com> ha scritto:

brunotl:
we can use 0.84 for min value, that require only 2 char change....
if you have spend time for check it work, it can be good.

Issue #1000, opened 2017-02-05
Select Waypoint : landable picto broken


brunotl:
![image](https://cloud.githubusercontent.com/assets/858626/22626773/b58de606-ebb5-11e6-8012-9eac2c00c548.png)

PR #1001, opened 2017-02-05
workaround issue #1000

brunotl:
picto size only depends of target rectangle.


pventafridda:
Non need to change anything in drawrunway, and no need to change the dlgWay.. It is only the scalefact passed by calling function which is wrong because the drawrunway changed  sizes after retuning. I fix it.

pventafridda:
NO. Sorry. I am wrong
Merging

PR #1033, opened 2017-02-19
rename FLARM to Traffic on map,

AlphaLima:
since FLARM is no longer the only traffic source.


pventafridda:
No, if you want to change the name you need to add a new token otherwise the old one will be untraslated for a long time and the whole thing useless. As a rule, never change tokens. And then you need to make it translated, and for 7 years we had this name on the page. So I am not pro-change.

pventafridda:
Ok I merge and fix it myself by adding new token.

pventafridda:
Ulli where is M280 used?? I can no longer find it

pventafridda:
It is not a radar, we only see Traffic on map, so Traffic was correct

Issue #1025, opened 2017-02-15
Screen orientation on Android


ToninoTarsi:
**New feature for Android** : At the moment screen orientation for Android is only based on internal sensor.
Especially for PG, with the instrument horizontally on the cockpit, sometime screen turns during take-off  ( it happened to me las t  Saturday during take-off ). Not all phone ( like mine ) can fix rotation in landscape mode. .. so ... something is missing.
Code ( both cpp and java ) is ready for implement  this but we do not have a setting for it ... 
Where to put it ? System config page 10 ? Screen modes ? ...

pventafridda:
system configuration I would suggest.


In data 15 Febbraio 2017 11:23:02 Tonino Tarsi <notifications@github.com> 
ha scritto:

pventafridda:
maybe it is better to make it configurable like this:
screen rotation
always enabled
enabled only in SIM mode
always disabled



In data 15 Febbraio 2017 11:23:02 Tonino Tarsi <notifications@github.com> 
ha scritto:

pventafridda:
but I wonder how we can rotate to landscape if it is disabled.


In data 15 Febbraio 2017 11:23:02 Tonino Tarsi <notifications@github.com> 
ha scritto:

ToninoTarsi:
System lock is disabled..  not app lock.
For option in current code we have   :
Automatic (current ad default ) 
Landscape
Portrait
Inverted landscape 
Inverted potrait (not working in all systems)

pventafridda:
ok great

ToninoTarsi:
Ok. Next week I can work on it adding the option to system menu 10

Sent from mobile

Il 15 feb 2017 4:54 PM, "Paolo Ventafridda" <notifications@github.com> ha
scritto:

ToninoTarsi:
Closed with merge #1034

PR #1034, opened 2017-02-19
Implemented Android Screen Orientation

ToninoTarsi:
Fix issue #1025


pventafridda:
Screen orientation must go in menu 22, System, not in User interface. And beside, there is no much space in menu 10 (though we should move Iphone gestures and Menu timeout to Engineering).

ToninoTarsi:
Moved to menu 22.

PR #1038, opened 2017-02-21
removed Odometer check, CALCULATED_INFO is a shared ressource.

AlphaLima:
However FAI Sectors<10km aren't drawn anyway now


pventafridda:
i have lost my head in this 6-stage single commit...!!

PR #1044, opened 2017-02-24
Added Black Glider Symbol

ToninoTarsi:
Added two more aircraft symbols completely black that solve a visibility problem on kobo.


pventafridda:
Remember.. the ToDo... the translators..

PR #1081, opened 2017-03-31
Glider sector airspace

AlphaLima:
added missing "glider sector" airspace
defined in Open Air as AC GSEC
defined in OpenAIP as GLIDER or GSEC

GSEC   is used in German official authority airspace 2017 as can be found here:
http://www.daec.de/fachbereiche/luftraum-flugbetrieb/luftraumdaten/

Change request here:
http://www.postfrontal.com/forum/topic.asp?TOPIC_ID=8533

### DIFF
--- a/Common/Distribution/LK8000/_System/CREDITS.TXT
+++ b/Common/Distribution/LK8000/_System/CREDITS.TXT
@@ -3,6 +3,7 @@
 [Latest software changes]
 In summary what's up in current branch:
 
+Airspace Class Glider Sector "GldSec" added (GSEC open airspace, GLIDING OpenAIP)
 Up to 6 devices can now be set in Device configuration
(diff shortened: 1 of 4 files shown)


pventafridda:
Ah! So finally you lazy boys do update the INFO!!!!  5 years I have been waiting for this moment!!

Common/Distribution/LK8000/_System/CREDITS.TXT
```diff
@@ -3,6 +3,7 @@
 [Latest software changes]
 In summary what's up in current branch:
 
+Airspace Class Glider Sector "GldSec" added (GSEC open airspace, GLIDING OpenAIP)
```
pventafridda: You also need to change the first line of the file, otherwise it wont be triggered on first run..

Issue #1088, opened 2017-04-09
Assertion failure on Kobo


pventafridda:
6.1F
KoboMini+BFV ver 10. Error (Update from version 60b)

After startup and on the waiting GPS fix I am get one error and only I can restart device.

CRITICAL ASSERTION FAILURE !
Execution failure in file
Common/Source/device.cpp
at line 65

LK8000 terminated !

pventafridda:
Is this now closed after your latest fixes, Bruno?

brunotl:
no, i have only added more diagnostic, for know where we use LockComm inside RxThread.

pventafridda:
So we need to check this out before releasing 6.1g for kobo

PR #1095, opened 2017-04-24
PExtractParameter : add size destination buffer parameter

brunotl:
avoid buffer overflow.

### DIFF
--- a/Common/Source/Library/StringFunctions.cpp
+++ b/Common/Source/Library/StringFunctions.cpp
@@ -16,44 +16,31 @@
 #include "Poco/TextConverter.h"
 #include "Util/Clamp.hpp"
 
-#ifdef __MINGW32__
-#ifndef max
-#define max(x, y)   (x > y ? x : y)
-#define min(x, y)   (x < y ? x : y)
-#endif
-#endif
-
-
-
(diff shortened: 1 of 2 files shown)


Common/Source/Library/StringFunctions.cpp
```diff
@@ -16,44 +16,31 @@
...
+  if ( CurrentFieldNumber == DesiredFieldNumber ) {
+    while( (dest_index < dest_size) && (index < StringLength) &&
+            (Source[ index ] != ',') && (Source[ index ] != 0x00) )
 	{
 	  Destination[dest_index] = Source[ index ];
 	  index++; dest_index++;
 	}
-      Destination[dest_index] = '\0';
-    }
+    assert(dest_index < dest_size);
```
pventafridda: Two cases. We use assert and remove LKASSERT or we use LKASSERT that by the way becomes automatically an assert when using DEBUG in makefile. Just asking because all reports from users, and I mean 100%, are about LKASSERT. No user can report about an assert error. Not to mention Windows CE...

brunotl: if LKASSERT are usefull, you can fix issues #1088

PR #1113, opened 2017-06-06
Bugfix

brunotl:


pventafridda:
If I merge this change, PNA and PC versions will not work as in the last 7 years...

brunotl:
ok, no problem, this bug is not critical, so we can leave it open ...

pventafridda:
If this commit is fixing on linux platforms, we should only make an IFDEF in order not to spoil the Windows platforms , that’s all.
Because I verified that on Windows profiles are working correctly: i wrote them long time ago for windows. After the change, they cant work anymore.
But at the same time, I verified that profiles on android dont work as on windows.

brunotl:
in fact i have missed to test "Save AS", now all work like it should on all
platform.

just for let you know : i have found the problem and fixed all this problem
on PC, and tested the fix on PNA just to be sure...

PR #1133, opened 2017-08-25
Fix

brunotl:
cf commit.


pventafridda:
ok maybe it is a good idea to disable BUGSTOP permanently from now on. We 
have a case of assert on mccready.cpp line 99.



In data 25 Agosto 2017 8:03:58 PM Bruno de Lacheisserie 
<notifications@github.com> ha scritto:

brunotl:
ok, but why this assert happen ?

pventafridda:
bestld is used by other threads.  That's why i used tmp variable.  If this is still true, the fix is introducing severe problems.

PR #1138, opened 2017-09-26
fix icon.png (file icon.png is not an actual png file.)

lshachar:
The file LK8000\android-studio\app\src\main\res\drawable\icon.png isn't actually a png file, but a link to lkicon.png.
This causes an exception while trying to compile lk8000 for android (armv7, fwiw):
Error:java.util.concurrent.ExecutionException: com.android.ide.common.process.ProcessException:

to solve, I copied LK8000\Common\Data\Bitmaps\LKicon.png over said copy of icon.png


ToninoTarsi:
To correctly compile on windows system with Android Studio you need to enable symlink on your system. Google " symlink  on windows" on how to do it.
I suggest to close this as the topic is also present in our Wiki : https://github.com/LK8000/LK8000/wiki/Android

lshachar:
Can't get it to compile properly, even though I have symlink enabled on my system & specific user. For some reason the file is not treated as a symlink on my system. but hey, if it works for everyone else..

ToninoTarsi:
Just try to clone with admin rights and symlink enabled on the register ( http://answers.perforce.com/articles/KB/3472/?q=enabling&l=en_US&fs=Search&pn=1 )

PR #1141, opened 2017-10-05
LiveTrack24 Radar - wrong help message in portrait mode

lshachar:
wrong help message in portrait mode (tiny correction)


ToninoTarsi:
That's correct. Should be merged. 
Thanks

PR #1143, opened 2017-10-18
fix: compare float to double are almost always true ...

brunotl:


pventafridda:
very important fix for performance thx

PR #1157, opened 2017-12-09
Automatically send Takeoff / landing SMS messages, user configurable.

lshachar:
Hi Guys,
## Intro:
 I'm creating this pull request even though that the code cannot currently be merged, and there are several things that I have not finished working on. (code does work, and I have been using it for several months now and I love it!)
Reason being, I'm going to be completely caught up with work until March, probably later, and maybe someone would like to pick it up and carry on where I had to stop. also, heck, I don't know if the core developers of LK8000 would like to include this feature in the software at all. and it's working well enough for me.
(I do know however that many other paragliders are very interested and waiting for this to be released)

## Alright.
So the point being, I wanted lk8000 on my smartphone to automatically send an SMS when I take off, saving me the hassle of letting my girlfriend / buddies at home know that I'm airborne (which I have done so far by manually texting them).
Then, they can follow my flight on LiveTracker24 (but they don't have to).
When LK8000 spots a landing, a messagebox is displayed. asking the pilot "Safe landing?"
Answering "Yes" sends a message to your contacts "I just safetly landed" (you can change the message, or you can define that safe landing messages won't be sent.)
if pilot clicked "No" or waited half the configurable time (say, it's set to 10 minutes, then, after 5 minutes) another message pops up: "send emergency message?"
answering with "YES" or waiting the rest of the configurable time, will send an emergency message. "Emergency landing. please help".
(IIRC, answering "No" will currently send no text messages)
audible alerts are played every 20 seconds, iirc. (not user configurable)
I added another screen to the Pilot Setup menu (Pilot setup now has Prev and Next buttons) where you set up the automatic messages. this should only appear for the Android target. (in fact non of the code is relevant to the other platforms)

## Screenshots:
2017-12-09 02 27 17 custom
2017-12-09 02 25 27 custom
2017-12-09 02 26 49 custom
2017-12-09 02 26 56 custom
2017-12-09 02 27 00 custom

## Todos:
**todo:** Pilot Setup menu: won't work in Longitude mode. I only added the new menu options to dlgConfigPilot_P.xml , so it outlines well how it should be added to dlgConfigPilot_L.xml.
**todo:** not all user configuration are saved. (This is what I was working on when I stopped having spare time to finish my work.) the message text is being saved (to the pilot configuration file) but the three on/off toggles to set what message is to be sent (1.takeoff message, 2.safe landing message, 3.emergency landing message) do not save. also the timeout value (to send the emergency message out) is not saved.
(look for "todo:sms:this is where I stopped" in LKProfileLoad.cpp)
**todo:** user configuration to automatically add a link to your LT24 page on all text messages. (Writing down the URL yourself on LK8000 keyboard is impossible, as you cannot type ":" or "/", so currently the user needs to edit their pilot setup configuration file, and manually add the URL there.)
**todo:** have more special characters in the keyboard layout file. one should be able to type these characters directly in LK8000 " : , ! ? ( ) "
**todo:** whatsapp / telegram integration. would save the text messages cost, can notify a group of people, and will not hickup on a long phone number list like your phone carrier would. (they might stop you from sending text messages after about 10 consecutive messages, for a certain period of time)
**todo:** resend failed text messages. (currently it only tries once. if you had no cellular reception LK8000 will never try re-sending the text messages. I think this should go in the Java code without the cpp code ever knowing if message was successfuly sent or not)
**todo:** in the java code, the "phoneNumbers.split(",");" function, was not happy splitting phone numbers with a dot "." so as a temporary fix I used a comma instead. that's an issue because LK8000 numerical keyboard does not contain a comma, so you can't type it to seperate numbers. there is a however a dot, so this should be looked into.
**todo:** text in messageboxes is hardcoded. It should be in the text files.
**todo:** my help file explanation text seems to be too long for lk8000... either find a way to shorten the text down or enlarge the buffersize for the help file text. (I suppose that's the issue)
**todo:** clean out my comments, within them are my friends phone numbers and lt24 user URL... remove some hardcoded variables which should not be hard coded (sms.cpp)
 :) can't even do THAT right now!

## future thought.
Idealy, there is still one case that I'd be happy to address one day:
If you had a bad landing, and you're out of phone coverage. (or phone's dead.) No text message would be sent. (People that received a takeoff message from you, may or may not notice that you have not sent a safe landing message, after a while)
a solution could be: a server (such as lt24) constantly gets updated with your whereabouts. when it is not updated for a certain time it notifies your friends "hey, THIS guy hasn't been seen since THIS time, and it was last seen HERE, doing THIS speed in THIS altitude". or something.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -87,6 +87,7 @@ add_library( LK8000 SHARED
     Common/Source/xcs/Android/EventBridge.cpp
     Common/Source/xcs/Android/InternalSensors.cpp
     Common/Source/xcs/Android/SoundUtil.cpp
+    Common/Source/xcs/Android/SmsUtil.cpp
     Common/Source/xcs/Android/BluetoothHelper.cpp
     Common/Source/xcs/Android/NativeLeScanCallback.cpp
     Common/Source/xcs/Android/PortBridge.cpp
@@ -779,6 +780,7 @@ add_library( LK8000 SHARED
     Common/Source/Library/glutess/tessmono.c
 
     Common/Source/Sound/Android/Sound.cpp
+    Common/Source/Sms/Android/Sms.cpp
 
     Common/Source/Resource/resource_data.S
     Common/Source/Resource/resource_xml.S
--- a/Common/Data/Dialogs/dlgConfigPilot_L.xml
+++ b/Common/Data/Dialogs/dlgConfigPilot_L.xml
@@ -1,7 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <PMML version="3.0" xmlns="http://www.dmg.org/PMML-3-0" xmlns:xsi="http://www.w3.org/2001/XMLSchema_instance" xsi:noNamespaceSchemaLocation="Dialog.xsd">
   <WndForm Name="frmConfiguration" Type="Dialog" X="0" Y="0" Width="320" Height="240" Caption="_@M195_" Font="2" >
-    <WndButton Name="cmdClose" Caption="_@M186_" X="3" Y="178" Width="62"  Height="42" Font="2" Tag="3" />
+    <WndButton Name="cmdNext"  Caption="_@M460_" X="3" Y="92" Width="62"  Height="42" Font="2" OnClickNotify="OnNextClicked" Tag="1" />
+    <WndButton Name="cmdPrev"  Caption="_@M6_" X="3" Y="-992" Width="62"  Height="42" Font="2" OnClickNotify="OnPrevClicked" Tag="0"/>
+	<WndButton Name="cmdClose" Caption="_@M186_" X="3" Y="178" Width="62"  Height="42" Font="2" Tag="3" />
     <WndFrame Name="frmPilot" X="70" Y="10" Width="-1" Height="222" Font="2">
       <WndButton Name="cmdPilotName" Caption="_@M525_" X="2" Y="1" Width="-1"  Height="35" Font="2" />
       <WndProperty Name="prpLiveTrackerInterval" Caption="_@M1779_" X="2" Y="70" Width="-1" Height="22" CaptionWidth="150" Font="2" Help="_@H1258_">
@@ -18,5 +20,7 @@
       <WndButton Name="cmdLiveTrackerusr" Caption="_@M1780_" X="2" Y="-1" Width="117"  Height="25" Font="2" />
       <WndButton Name="cmdLiveTrackerpwd" Caption="_@M1781_" X="120" Y="-999" Width="-1"  Height="25" Font="2" />
     </WndFrame>
+  
+  <!-- TODO: Add sms message configuration screen, from dlgConfigPilot_P -->
   </WndForm>
 </PMML>
--- a/Common/Data/Dialogs/dlgConfigPilot_P.xml
+++ b/Common/Data/Dialogs/dlgConfigPilot_P.xml
@@ -1,22 +1,55 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <PMML version="3.0" xmlns="http://www.dmg.org/PMML-3-0" xmlns:xsi="http://www.w3.org/2001/XMLSchema_instance" xsi:noNamespaceSchemaLocation="Dialog.xsd">
   <WndForm Name="frmConfiguration" Type="Dialog" X="0" Y="0" Width="240" Height="320" Caption="_@M195_" Font="2" >
-    <WndButton Name="cmdClose" Caption="_@M186_" X="5" Y="270" Width="105"  Height="30" Font="2" Tag="2" />
+    <WndButton Name="cmdNext"  Caption="_@M460_" X="115" Y="238" Width="105"  Height="30" Font="2" OnClickNotify="OnNextClicked" Tag="1" />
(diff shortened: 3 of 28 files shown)


brunotl:
very nice feature ! thanks

pventafridda:
I suppose it is for next version after 6.2... right?

pventafridda:
Ok I have quickly looked at the commit. There are several issue with this commit.
The first is: it is an android dedicated function, cannot work on microsoft platforms, raspberry , kobo etc. 
The second is: this is a SAR (search and rescue) aid function. But it does not work the way such a function should: the pilot has absolutely no idea if the SMS has gone through, and LK will not resend the SMS once under coverage. 
The third, and most important, is: while in troubles, on a glider, I wouldnt even consider spending 5 seconds to send an SMS . But this is only a personal opinion (as a mountain pilot, we all know that if there is a chance to land safely every second counts, if there is no chance and it is going to be a crash landing, every second counts twice). 
Much more useful would be to send out a position report sms, similar to what we do with emergency radio report for paragliders now.

Anyway, the issue about granting delivery of SMS as soon as we have network coverage is not optional. Without it, the whole idea is worthless in my opinion.
I have serious doubts about this merge, right now, despite the fact it is a great add-on and we also gain a new developer!

ToninoTarsi:
I have this future in my [LiveTrack24 App](https://play.google.com/store/apps/details?id=com.tony.ttlivetrack24v2) and it is very useful  and would be great to have it also in LK8000. 
I my App I decided not to give the opportunity of sending an "emergency SMS" but only "start" and "stop". This for the same reason Paolo pointed out above. But as an option could be useful for someone.
I did not look at the code but a great effort shroud be done to assure the message is correctly managed in difficult network coverage.

pventafridda:
I agree. First we should set the problem of sms being accepted by the 
network, if possible. If this is possible then we can also provide 
emergency sms. If not, we simply give the user takeoff and landing sms.
I don't think it is necessary to an age a spooler and resend sms. But at 
least knowing if the sms has gone through is imperative for the emergency 
message.

Apart from this, we may in the future as well provide position reporting to 
other friends by sending our position and processing it in parallel of the 
current friend codes.


In data 11 Dicembre 2017 14:28:20 Tonino Tarsi <notifications@github.com> 
ha scritto:

lshachar:
Hey guys,
Thanks for reviewing my work!
[It is completely possible to get listeners for both delivery and sending of text messages](https://stackoverflow.com/questions/18771356/check-if-an-sms-is-actually-sent), and so making sure that the messages did went through.

([it is also possible to check the signal strength prior to sending](https://stackoverflow.com/questions/13275663/check-network-coverage-in-android-before-trying-to-send-sms))

Paolo:
> The first is: it is an android dedicated function, cannot work on microsoft platforms, raspberry , kobo etc.

Yeah, currently that is the case. but later we can add support for other devices that are able to send text messages.
> The second is: this is a SAR (search and rescue) aid function. But it does not work the way such a function should: the pilot has absolutely no idea if the SMS has gone through, and LK will not resend the SMS once under coverage.

(Once we finished the code that makes sure that the message was indeed sent successfully) There is already a status message box that tells the pilot that a text message was sent. We can change it to a messageboxX that the pilot must click Okay to acknowledge.

> The third, and most important, is: while in troubles, on a glider, I wouldnt even consider spending 5 seconds to send an SMS . 

Well that is _exactly the point!_ If you don't do anything - the emergency message is triggered!
IE, If you crash badly enough that you are immobile, LK first registers a landing, and after the emergency timeout runs out - the emergency messages are automatically sent.
You should not (and in fact cannot) do anything except:
1.set it up to your liking (send takeoff message? send safe landing? or only send emergency landing?)
2.After landing, tell LK if you safely landed.

pventafridda:
What happens if pilots set emergency landing message, they land safely as 
usual on their airport and forget to acknowledge? Message is sent. So we 
send out SAR messages without a reason. All gliders already have a 
localisation device for SAR that activates after impact, it does not need 
any phone network and it is traced also by sats. It is mandatory to use it 
aboard.

Anyway takeoff and landing messages can be used without any problem but 
without SAR purposes.




In data 26 Dicembre 2017 08:52:18 lshachar <notifications@github.com> ha 
scritto:

lshachar:
I wish paragliders and hang gliders had the same instruments that are mandatory in gliders. Unfortunately that is not the case. When a paraglider has a serious accident - nothing happens. If the pilot is fortunate enough to have someone worry for him - then maybe within a few hours someone acknowledges that he's missing.
FYI, once the landing confirmation message appears, it plays a sound every 20 seconds.
When 1/2 the timeout time had passed, and the second message appears - a much longer, louder sound plays. (LK_BOOSTER - the worse sound effect I could find!)
-It did happen to me already that the sound was what drew my attention to my phone - to click the safe landing confirmation screen.

> What happens if pilots set emergency landing message, they land safely as usual on their airport and forget to acknowledge? Message is sent. So we send out SAR messages without a reason.

True. but, is that truly the **worse-case** situation? IE, my dad and my girlfriend will call me on my cellphone, and I'll apologize to them, tell them that I have landed safely and all is well.
both me & them are happy to pay that price for the peace of mind this feature normally offers.

(also, clearly you are talking about gliders (which land in airports, vs fields for the rest of us,) and I don't see why a glider pilot would use the emergency landing message if they already has a proper SAT SAR device.)

pventafridda:
Yes you are right for paragliding it makes a lot of sense. We have to work 
on merging next to 6.2


In data 26 Dicembre 2017 3:23:09 PM lshachar <notifications@github.com> ha 
scritto:

PR #1164, opened 2018-01-10
fix rounding glitch for brightness and contrast in config dialog

AlphaLima:
fixed this:
http://www.postfrontal.com/forum/topic.asp?TOPIC_ID=8718


brunotl:
why removing iround() call ?

pventafridda:
TerrainBrightness is a SHORT, not a float. We ought to use iround.
The problem you try to fix may be instead related to the SetAsFloat.

AlphaLima:
I don't mind. I don't want to investigate the real cause of it and I can live with the glitch.

AlphaLima:
OK, made it with iround!

pventafridda:
I merge it now. Ulli please ensure it works properly!

PR #1166, opened 2018-01-29
Waypoint Dialogs Caption Info

AlphaLima:
Waypoint Dialog Caption Info (frequency + runway) for waypoint not only for CUP file data.
For OpenAIP waypoint files, Frequency and raunway information was excluded for dialog caption


pventafridda:
Only CUP file (and AIP) are doing a correct and managed init of  WayPointList  Code, Freq etc.
So we should add OpenAIP file type aside CUP, and not consider DAT, OZI, COMPE, etc.
I cannot merge it like this.

AlphaLima:
OK done!

PR #1169, opened 2018-02-20
new Airspace lookup dialog with

AlphaLima:
new Airspace lookup dialog like waypoints with
-substring filter 
-pictorial
-easy turn off/on
-new Filter for disabled airspaces
-new Upper-/lowercase keyboard layout

comming from this request:
http://www.postfrontal.com/forum/topic.asp?TOPIC_ID=8732
for dense airspaces like netherland really needed!

result:
Note the Altdorf is disabled but can be re-enabled by just one click in the listfield. Both elements of the same airspace are handeled in parallel (due to group acknolage).
notam_alt2

sorting out NOTAMs in max. 50km distance
notamnew

sorting out the disabled airspaces for easy re-enabling
disable

new upper-/lowercase switchable keyboard for all Text inputs
lowercase


AlphaLima:
the new keyboard works for all text edits.
agree should be tw commits, but the airspace lookup can not work propperly without the new keyboard.
(BTW: same for terminal send)

brunotl:
1 - do git rebase on top of master for solve conflict
2 - make test on utf8 plaform (android or linux) with airspaces name containing multibyte code points

brunotl:
replaced by #1219

PR #1170, opened 2018-02-23
Airspace Lookup button

AlphaLima:
new Meneu button in NAV 3/3 Menue for the new Airspace Lookup dialog.


Common/Source/LKInterface/LKCustomKeyHandler.cpp
```diff
@@ -320,7 +320,10 @@ bool CustomKeyHandler(const int key) {
 		PlayResource(TEXT("IDR_WAV_CLICK"));
 		InputEvents::eventWaypointDetails(_T("select"));
 		return true;
-
+    case ckAirspaceLookup:
```
brunotl: case must be ordered

Common/Source/LKInterface/LKCustomKeyHandler.cpp
```diff
@@ -320,7 +320,10 @@ bool CustomKeyHandler(const int key) {
...
+        PlayResource(TEXT("IDR_WAV_CLICK"));
```
brunotl: why not the same sound ? cf. InputsEvents.cpp line 2019

Common/Source/InputEvents.cpp
```diff
@@ -2105,6 +2105,13 @@ void InputEvents::eventService(const TCHAR *misc) {
 	return;
   }
 
+  if (_tcscmp(misc, TEXT("AIRSPACELU")) == 0) {
+      LKSound(TEXT("LK_KLICK.WAV"));
+       dlgAirspaceSelect();
+        return;
+  }
+
```
brunotl: why not use CustomKeyHandler ? 
like that you avoid to have two different code for same feature ....1

Common/Source/LKInterface/LKCustomKeyHandler.cpp
brunotl: case must be ordered

Common/Source/LKInterface/LKCustomKeyHandler.cpp
brunotl: why not the same sound ? cf. InputsEvents.cpp line 2019

Common/Source/InputEvents.cpp
brunotl: why not use CustomKeyHandler ?
like that you avoid to have two different code for same feature ....1

Common/Header/LKInterface.h
```diff
@@ -103,6 +103,7 @@ typedef enum{
 	ckResetComm,
 	ckDspMode,
 	ckOtimizerPointsToggle,
+	ckAirspaceLookup,
```
brunotl: if you add CustomKey, you need also add CustomKeyLabel

Common/Source/LKInterface/LKCustomKeyHandler.cpp
AlphaLima: ???

Common/Source/LKInterface/LKCustomKeyHandler.cpp
AlphaLima: Just copied it from the case above, why make it different than other menue entries?

Common/Source/LKInterface/LKCustomKeyHandler.cpp
brunotl: writing

switch(..) {
case 1:
case 2:
case 3:
}

are better than writing

switch(..) {
case 3:
case 1:
case 2:
}

especially when all cases are contiguous value

Common/Source/LKInterface/LKCustomKeyHandler.cpp
brunotl: ok, here it's not menu entrie, it's custom key... 
so if i configure custom with this, i don't have same sound than standart menu you have added ...

after that, agree, this is probably the right sound, for avoid this problem, you can change your change in InputEvents.cpp using same way used for "SONAR" or "OPTIMODE"

Common/Header/LKInterface.h
AlphaLima: done!

PR #1171, opened 2018-02-24
OpenAIP airspaces

AlphaLima:
OpenAIP airspaces were skipped on unknown type or if no name. Not a good idea, possible loss of important informations.
This patch keeps unknown type airspaces as type other/unkown
If an airspace has no name the name is set to "noname"

### DIFF
--- a/Common/Source/Airspace/LKAirspace.cpp
+++ b/Common/Source/Airspace/LKAirspace.cpp
@@ -1987,8 +1987,8 @@ bool CAirspaceManager::FillAirspacesFromOpenAIP(ZZIP_FILE *fp) {
         // Airspace category
         dataStr=ASPnode.getAttribute(TEXT("CATEGORY"));
         if(dataStr==nullptr) {
-            StartupStore(TEXT(".. Skipping ASP with no CATEGORY attribute.%s"), NEWLINE);
-            continue;
+            StartupStore(TEXT("..  ASP with no CATEGORY attribute.%s"), NEWLINE);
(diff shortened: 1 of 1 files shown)


pventafridda:
I merge it given the importance and patch simplicity, but please ensure that even with wrong airspace files the "other" type will not lead to crashes!

Common/Source/Airspace/LKAirspace.cpp
```diff
@@ -1987,8 +1987,8 @@ bool CAirspaceManager::FillAirspacesFromOpenAIP(ZZIP_FILE *fp) {
         // Airspace category
         dataStr=ASPnode.getAttribute(TEXT("CATEGORY"));
         if(dataStr==nullptr) {
-            StartupStore(TEXT(".. Skipping ASP with no CATEGORY attribute.%s"), NEWLINE);
-            continue;
+            StartupStore(TEXT("..  ASP with no CATEGORY attribute.%s"), NEWLINE);
+         //   Type = OTHER;
         }
         size_t len=_tcslen(dataStr);
```
brunotl: What Happen if dataStr == nullptr  ( cf. line 1989 ) ?

pventafridda: should be type other and continue.
nota a good idea to merge last minute commits.. should we reverse?

brunotl: _tcslen(nullptr)  -> segfault

PR #1172, opened 2018-02-24
Atr833fix

AlphaLima:
The ATR833 did work unidirectional only.
This fix send the required "KeepAlive" Sentence every 5s.
Timer derived from Update devices


pventafridda:
I would leave it for next version ..

AlphaLima:
Done!

PR #1173, opened 2018-02-26
check for NULL Pointer added

AlphaLima:
Sorry guys seems I did not undo all stupid changes I tried out.
The blamed line originally was not intended to be committed.
Now also with better null pointer checking


brunotl:
we can' merge that as is ...
you need to  rebase and make clean  commit history...

AlphaLima:
replace by another commit

PR #1176, opened 2018-03-03
Support unknown and unnamed OpenAIP airspaces

AlphaLima:


AlphaLima:
>naming all of them with same name is bad idea because if "group aispaces" option are enable all unamed >airspaces are grouped ....

intended, so we can disable them all on one click. If a user want to handle it seperatly, give it a name!

>with your change some unamed aispaces still skipped
where?

brunotl:
> > naming all of them with same name is bad idea because if "group aispaces" option are enable all unamed airspaces are grouped ....
> 
> intended, so we can disable them all on one click (still beter than just drop it without warning). If a user want to handle it seperatly, give it a name!
> we can auto name it like "Unkwn1, Unkwn2, Unkwn3, ..." but I don't thik this is what we like to have

unamed airspace is bug in airspace files, so i think it's better to not group them

> 
> > with your change some unamed aispaces still skipped
> 
> where?

cf. line 2071 and 2075 of your changes

AlphaLima:
>unamed airspace is bug in airspace files, so i think it's better to not group them
True, its a bug in the airspace file. I don't mind if these "bugs" can be grouped or not. I want to know if there is a potential airspace I might be aware of (if it is important). Up to know it is simply ignored without any aknolagement and that is ever worse!
 
>cf. line 2071 and 2075 of your changes
2071 if just a Null pointer check, like many other places of that file.
2075 is for empty nodes only, which does not define an airspace at all

PR #1178, opened 2018-03-09
All tasks from CUP Files

AlphaLima:
now open a Task select box if more than one Task defined in a CUP file.
multitasks


brunotl:
- wall file are modified, so impossible to solve conflict...
- "dialog"  must be moved in it's own files

brunotl:
replaced by #1221

PR #1182, opened 2018-04-18
Flarm declaration rework

AlphaLima:
On NMEA Multiplexed systems the Flarm declaration failed due to interfering NMEA sentences sent by other devices.
This Fix solves this problem by sorting out the correct Flarm NMEA Answers from the others with a timeout of 500ms


brunotl:
you have change ExpectString() function, this fonction are also used by some other devices.
what is the side effect of this change on this other devices ?

AlphaLima:
I don't think it will have sideeffects but I'm not 100% sure 
so I made it exclusive for flarm in an own function.

AlphaLima:
delete unneeded files

PR #1188, opened 2018-05-13
Exclude Airspaces outside Terrain

AlphaLima:
When waypoint filter is set to exclude now airspaces outside Terrain (OpenAir and OpenAIP) are also excluded.
Maybe we should rename "Exlude Wpt" to "Exlude Data"
The terrain filter area is a bit extended by 5% in each direction to guarantee all airspaces "on the edges" not to be removed.

Very helpfull on bigger countries with lots of airspaces. It gives a bigger performance by removing the uninteresting airspaces outside the dedicated flying area/terrain as implemented here:
https://github.com/LK8000/LK8000/pull/1191

Terrain Filter include (as before):
terrainfilter_off

Terrain Filter exclude:
terrainfilter_on

### DIFF
--- a/Common/Source/Airspace/LKAirspace.cpp
+++ b/Common/Source/Airspace/LKAirspace.cpp
@@ -1544,7 +1544,7 @@ void CAirspaceManager::FillAirspacesFromOpenAir(ZZIP_FILE *fp) {
     double lat = 0, lon = 0;
     bool flyzone = false;
     short maxwarning=3; // max number of warnings to confirm, then automatic confirmation
-
+    bool InsideMap = false;
     StartupStore(TEXT(". Reading OpenAir airspace file%s"), NEWLINE);
     charset cs = charset::unknown;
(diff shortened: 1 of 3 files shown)


Common/Source/Airspace/LKAirspace.cpp
```diff
@@ -1603,13 +1603,15 @@ void CAirspaceManager::FillAirspacesFromOpenAir(ZZIP_FILE *fp) {
                                     }
                                 }
                             }
-                            if (newairspace) {
+                            if(InsideMap) {
```
brunotl: if this condition is false, "newairspace" created line 1602 or 1596 is not added to _airspaces and pointer are set to null line 1629, and never deleted, so you have memory leak ...

AlphaLima:
already solved, created a complete new pull request.
This one can be closed!

brunotl:
Replaced by #1211

PR #1189, opened 2018-05-13
fix LED and "Disabled" Problem in Device Dialog

AlphaLima:


AlphaLima:
Important fix for device setup

brunotl:
I don't what is fixed because that work better without fix ...

AlphaLima:
-MAXNUMDEVICES removed because we already have the same definition NUMDEV.
-When more than two devices are enabled (A, B and F) and we restart LK Device B was disabled.
-The LED indicators didn't show the enabled devices.

Common/Source/Dialogs/dlgConfiguration.cpp
```diff
@@ -4751,18 +4750,27 @@ void InitDlgDevice(WndForm *pWndForm) {
...
-  for(unsigned i = 0; i < MAXNUMDEVICES; ++i) {
+  for(unsigned i = 0; i < NUMDEV; i++) {
     TCHAR szWndName[5];
     _stprintf(szWndName, _T("cmd%c"), _T('A')+i);
     WindowControl * pWnd = pWndForm->FindByName(szWndName);
     if(pWnd) {
       pWnd->SetWidth(w);
       pWnd->SetLeft(lx);
+
+      if(_tcslen(dwDeviceName[i]) == 0)
```
brunotl: that's wrong, device name can be "DISABLED"

PR #1190, opened 2018-05-14
Send AutoMC changes to Devices 

AlphaLima:


brunotl:
don't work if device is not connected before "devPutMacCready()" first call.

AlphaLima:
If the device is not connected, it will try only once and on every change of the MC value only.
That is intended to prevent to many transmission try.
Do you want to send continously  to a non present device? Thats something Paolo does not like.

brunotl:
no need to send it continously, you just need to send when it's needed
 - when device is connected
 - when value change

AlphaLima:
Yes, and that is exactly how it works after my last commit!

brunotl:
only if you use LXV7_EXP device drivers, not for others ( cai302, LX16xx, LXminimap, LXV7,  openvario, Vaulter, Westerboer .... )

AlphaLima:
Why?

brunotl:
how change in "DevLXV7_EXP::ParseNMEA(...)" can change anything on other device ?

AlphaLima:
ParseNMEA has nothing to do whith setting MC on devices but read back the new setting only.
In this case (DevLXV7_EXP) it prevents circular updates between LK and the device by a update holdoff time.

brunotl:
after more check, this can't be merged as is because "CheckSetMACCREADY()" fonction are used by some device Drivers when MC value are get from device. 

this need to be refactored taking account the general case : 
   - device can send MC value to LK8000
   - LK8000 can send MC value to device

so 
   - who is the master ( device or LK8000 or none) ?
   - what happen if both send data to each others ?
   - what happen if two device connected send there MC Value ?

i close this pull request waiting for refactoring ...

PR #1193, opened 2018-05-27
In waypoint lookup dialog:

AlphaLima:
searching not only for waypoint name but also for waypoint code


AlphaLima:
found a better way, commited and opended another pull request!

PR #1195, opened 2018-06-09
Flarm dialog

AlphaLima:
reading IGC Files from a FLARM device:
For Flarm devices we have now a Config Button in the Divice setup. There we can download IGC Files from Flarm. The Files are stored in the _Logger directory.
workflow pictures:
image
image
image
image
image


AlphaLima:
replaced by:
https://github.com/LK8000/LK8000/pull/1272

PR #1199, opened 2018-07-03
Enable linux fdr

AlphaLima:
Enable Flight data Recorder FDR also for non Windows devices.


brunotl:
are you sure ?

brunotl:
- too much commit not related to Flight Data Recorder
- no need specific code for PNA
- gmtime() is not thread safe 

replaced by pull request #1200

PR #1200, opened 2018-07-04
Enable Flight Data Recorder on linux target

brunotl:
replace pull request #1199

### DIFF
--- a/Common/Header/options.h
+++ b/Common/Header/options.h
@@ -40,8 +40,6 @@
 #endif
 
 #ifdef __linux__
- // temporary disable, need to be port...
- #define NO_DATARECORDER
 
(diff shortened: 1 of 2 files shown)


AlphaLima:
Linux filenames are case sensitive, we open the Config File:
FlightRecorder.CFG
according to the documentation https://github.com/LK8000/LK8000/blob/master/Docs/FlightRecorderDoc.txt
 it should be
FlightRecorder.CFG
or even better both!
https://github.com/AlphaLima/LK8000/commit/7267ebb1b62b5501f270a24318b6d730a1653bbb

brunotl:
@AlphaLima : your commit is in pull request now

Issue #1180, opened 2018-03-26
AUTO Bottom bar mode


ToninoTarsi:
Hi all,

Just a small consideration about AUTO Bottom bar switching mode.

In the current behaviour if AUTO is selected in bottom bar configuration dialogue the TRM0 bottom bar will automatically be activated while thermalling and deactivated after exiting thermal.
That's wonderful for HangGliding and avoid instrument interaction for having always the current informations.
The only minor problem I face with it  is that, if I forget to to set the bottom bar in something different than TRM0 at take-off, the automatic switching does not work. 

Not a real problem indeed but I was just wandering if a behaviour based on MapWindow::Mode would be more correct : 
1) we have 3 major Modes : MODE_CRUISE,MODE_CIRCLING and MODE_FINAL_GLIDE
2) we have 3 major bottom bar : CRU7,TRM0 and FIN8

Why not to link this 2 point  feature in AUTO Bottom bar mode ?

Tony

ToninoTarsi:
Just to better understand what I mead this is test branch ( to be cleaned and completed )
https://github.com/LK8000/LK8000/compare/master...ToninoTarsi:BottomBarAuto
If you agree with this add-on I will finish it after 7 is out to be inserted in next beta.
Tony

PR #1202, opened 2018-07-09
Bottom bar auto

ToninoTarsi:
As discussed here : https://github.com/LK8000/LK8000/issues/1180


brunotl:
just a little comment : 
it can be better to use enum instead of  short + litteral for "ConfBB0Auto"

ToninoTarsi:
I opted for #define as it is the style used in all that part of the code.

brunotl:
sorry tony, problem come from merge odrer ...
can you solve conflict and update Language\Todo;txt with your change in language and help  files ?

Issue #1132, opened 2017-08-14
New waypoint DD.dddd, 9999 after inserting 5 numbers


parapenT1sta:
When creating a new waypoint using format DD.dddd, if we introduce 54354 for example, instead of assuming jus the 4 numbers it changes to 9999. 

Is anyway to limit to 4 numbers while inserting the number or just assume the first 4 in case of having 5 or 6 numbers?

parapenT1sta:
@brunotl @pventafridda @AlphaLima 

Anyone? 😄

parapenT1sta:
@AlphaLima Can you fix this?

AlphaLima:
https://github.com/LK8000/LK8000/pull/1205

parapenT1sta:
@AlphaLima Thank you!!!! :)

PR #1205, opened 2018-08-18
Fix the 9999 problem on waypoint edit in DDdddd Format

AlphaLima:
as complained here:
https://github.com/LK8000/LK8000/issues/1132#issuecomment-413801736


brunotl:
don't work if you enter "10000"

AlphaLima:
OK now!

PR #1207, opened 2018-09-11
Flarm detect

AlphaLima:
Flarm Traffic from OGN Viewer was wrongly positioned due to false initialized FLARM Factor by missing $PFLAU sentences.


AlphaLima:
obsolete! OGN viewer added PFLAU sentences.

PR #1209, opened 2018-10-04
Terminal Send

AlphaLima:
added send Line with text edit to Terminal window.
If entered text is a NMEA String (starting with $) the checksumm is added automatically.
This makes sense with the new keyboard only, otherwise we can not enter special caracters as needed for NMEA sentences.

terminal


AlphaLima:
any reason not to merge this pull request?

PR #1215, opened 2018-11-25
Overlay are now customizable directly in dlgOverlays dialog.

ToninoTarsi:
Removed the link from overlay customization and AUX infobox group. Overlay are now customizable directly in dlgOverlays dialog. To avoid any conflict the old configuration the customizable overlay have been renamed in configuration.


brunotl:
merged manualy : 0658d28279910f2b550225a4e987b385c64e2105

PR #1217, opened 2018-11-26
Overlay are now customizable directly in dlgOverlays dialog.

ToninoTarsi:
Removed the link from overlay customization and AUX infobox group. Overlay are now customizable directly in dlgOverlays dialog. User configuration is maintained.


brunotl:

PR #1228, opened 2018-12-20
Translate Android Key Codes into Unicode characters

TheEnlightenedFlorist:
In general, Android Key Codes should not be used directly. Change translateKeyCode in NativeView.java to translate the Android Key Codes into Unicode characters. EventBridge.cpp also has a translateKeyCode function. Whatever it was doing was wrong and it should now be unnecessary.

I've been creating a custom controller with Arduino and this was necessary to get the minimap to move right with the space key. I've tested all of the buttons on my controller and they work correctly now. There is a little bit of functionality that is still not working correctly, but this gets it 90% of the way there.

### DIFF
--- a/Common/Source/xcs/Android/EventBridge.cpp
+++ b/Common/Source/xcs/Android/EventBridge.cpp
@@ -29,6 +29,8 @@ Copyright_License {
 #include "Compiler.h"
 #include <jni.h>
 
+#define ANDROID_KEYCODE_FIX
+
 /**
  * @see http://developer.android.com/reference/android/view/KeyEvent.html
  */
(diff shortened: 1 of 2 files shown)


brunotl:
First, Welcome in contributor list !

two questions :
  - have you tested your changes with real keyboard ?
  - what is the "functionality still not working" ?

TheEnlightenedFlorist:
I don't have a real bluetooth keyboard to test with. They should all be the same if they conform to the HID protocol. Has anybody tested bluetooth keyboards on Android before? I suspect this may have been an issue with all of them.

The functionality that's not working mostly has to do with dialogs. I can't move the highlighted button or select the highlighted button. The code I modified seemed to only deal with inputs on the main page and I haven't looked into dialogs yet.

TheEnlightenedFlorist:
I got a USB OTG cable and tested it with my desktop keyboard. I experienced the same issues I had with my controller and my fixes had the same effect.

PR #1232, opened 2019-01-05
Multi select reworked

AlphaLima:


brunotl:
last commit added to late, not merged ....

PR #1234, opened 2019-01-06
renamed SFC in the more common phrase GND and moved it to language file.

AlphaLima:
just a small change for a very old cause of misunderstandings.

### DIFF
--- a/Common/Source/Airspace/LKAirspace.cpp
+++ b/Common/Source/Airspace/LKAirspace.cpp
@@ -3217,7 +3217,7 @@ void CAirspaceManager::GetAirspaceAltText(TCHAR *buffer, int bufferlen, const AI
             break;
         case abAGL:
             if (alt->AGL <= 0)
-                _sntprintf(intbuf,BUF_LEN,TEXT("%s"), MsgToken(2387)); //_@M2387_ "GND"
+            	CopyTruncateString(intbuf, BUF_LEN, MsgToken(2387)); // _@M2387_ "GND"
             else {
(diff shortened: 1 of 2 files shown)


Common/Source/Airspace/LKAirspace.cpp
```diff
@@ -3216,7 +3216,7 @@ void CAirspaceManager::GetAirspaceAltText(TCHAR *buffer, int bufferlen, const AI
             break;
         case abAGL:
             if (alt->AGL <= 0)
-                _stprintf(intbuf, TEXT("SFC"));
+                _stprintf(intbuf, MsgToken(2387)); //_@M2387_ "GND"
```
brunotl: sprintf with user format string : unsafe

Common/Source/Airspace/LKAirspace.cpp
```diff
@@ -3254,7 +3254,7 @@ void CAirspaceManager::GetSimpleAirspaceAltText(TCHAR *buffer, int bufferlen, co
             break;
         case abAGL:
             if (alt->AGL <= 0)
-                _stprintf(intbuf, TEXT("SFC"));
+                _stprintf(intbuf, MsgToken(2387)); // _@M2387_ "GND"
```
brunotl: sprintf with user format string : unsafe

Common/Source/Airspace/LKAirspace.cpp
AlphaLima: solved

Common/Source/Airspace/LKAirspace.cpp
AlphaLima: solved

AlphaLima:
done!

PR #1255, opened 2019-01-14
added missing text tokens, some corrections

AlphaLima:
added missing token for system menue 12 "Fonts"
some little corrections


Common/Data/Language/ENG_MSG.TXT
```diff
@@ -1031,7 +1031,7 @@ _@M975_ " GPS data error "
 #
 # V6.2
 #
-_@M981_ "Auto-Contrast"
+_@M981_ "Auto-Contrast "
```
brunotl: why added space ?

AlphaLima: because all in this dialog have it:
image

PR #1265, opened 2019-01-17
 less intrusive background android notification

brunotl:


brunotl:
according to documentation channel with low priority don't play sound
https://developer.android.com/training/notify-user/channels

PR #1274, opened 2019-01-22
IGC logger : Rewrite signing algorithm

brunotl:
- remove intermediate file
 - G Record updated and appended after each file update

data written to the IGC file only come from internal memory and file are always signed, even if software is not closed properly.


ToninoTarsi:
I do not see (or maybe I'm missing) any reference to a secret key required by FAI at point 3.1.4.3

PR #1275, opened 2019-01-22
Modified minimum reasonable AutoZoom limit for PG mode

ToninoTarsi:
In PG mode the minimum reasonable AutoZoom limit was only 10. This can lead to a very
confusing situation for the pilot that never have a clear filling on where the cylinder
is as the zoom continues to increase. This new hard coded value should me more reliable
but in the future a rework, eliminating also ISPARAGLIDER, will be necessary.


ToninoTarsi:
Replaced by #1284

Issue #1130, opened 2017-08-06
Change feature: Task Time for PG


parapenT1sta:
As far I can see, LK8000 start counting the task time, when you hit the start but I think it should take in count the Start Time set up on the PG Time Gates at least when the Time Gates is setup as 1.

parapenT1sta:
@brunotl Is it possible to be done?

ToninoTarsi:
Agree .. should be fixed !

parapenT1sta:
Thanks 👍

PR #1276, opened 2019-01-22
Task start time when a TimeGate is active

ToninoTarsi:
If a task start TimeGate is present task start time should be equal the the active TimeGate.
Resolves:  #1130


parapenT1sta:
Thank you so much!

PR #1278, opened 2019-01-23
Reset Task if user select to go back to Start

ToninoTarsi:
If the user manually select to go back to start and Time Gates are present Task should be restarted with current TimeGate.


ToninoTarsi:
This is not intended to be merged but just to start a discussion about.
Problem :  if pilot, after have crossed the start cylinder, decide to go back to start should the task be restarted? 
For Hangliding then several start gates are present I think maybe yes. And this is what PR does. Should it be the same for GL or for PG with only one start gate ?

ToninoTarsi:
Exact, So this PR does not affect Glider mode or PG mode if only one Start gate is present.
But imagine this scenario that is quite common in HG competitions :
- Pilots decided to make second ( or any > 1 ) start point.
- Waiting for second start he can cross the start several times. Also to test conditions on the way.
- I'm sure that the intuitive thing he will do to reset to start is to make a back that will not lead him to correct start gate ( I also was mislead by this )

On the other side what contraindication can this PR have ?  At the moment I do not see any but ...

brunotl:
a "contraindication" can be a user input mistake ... ( can be solved by asking to user if he want to reset )

ToninoTarsi:
Yes .. agree.  Adding a confirmation message will give the pilot the freedom to choose and will also make him aware of what is going on going back to start.
I will rework on that direction.

Common/Source/InputEvents.cpp
```diff
@@ -3397,13 +3397,22 @@ void NextUpDown(int UpDown)
           AdvanceArmed = false;
           ActiveTaskPoint = 0;
           CALCULATED_INFO.LegStartTime = GPS_INFO.Time ;
+          CALCULATED_INFO.LegStartTime = GPS_INFO.Time ;
```
brunotl: please remove this duplicated ligne

PR #1281, opened 2019-01-29
Enable external sound also on Android devices

ToninoTarsi:
Some Android devices (especially e-ink) does not have sound.
This enable external sound on all Android and so user has a choose to use it.


ToninoTarsi:
Not so strait forward as I thought

PR #1293, opened 2019-03-04
NANO3 IGC Download

AlphaLima:
added IGC Download Dialog to LXNav Nano3 Driver.
added Configuration Dialog, more elements will be added to become an universal LXNav driver


brunotl:
tha can't be merged as is, please do a correct rebasing

PR #1304, opened 2019-03-26
bugfix, recalc water ballast directly after ballast factor changed

AlphaLima:
prevent possible division by zero on gliders with no ballast
moved LX <-> LK  conversion functions to Utils.h as common shared code

### DIFF
--- a/Common/Source/Devices/devLX16xx.cpp
+++ b/Common/Source/Devices/devLX16xx.cpp
@@ -11,6 +11,7 @@
 #include "devLX16xx.h"
 #include "LKInterface.h"
 #include "Baro.h"
+#include "Utils.h"
 
 int iLX16xx_RxUpdateTime=0;
 double oldMC = MACCREADY;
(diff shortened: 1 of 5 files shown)


Common/Source/Devices/devLX16xx.cpp
```diff
@@ -356,12 +357,12 @@ bool DevLX16xx::LXWP0(PDeviceDescriptor_t d, const TCHAR* sentence, NMEA_INFO* i
     info->VarioAvailable = FALSE;
 
 
+
```
brunotl: Why ?

brunotl: ok, so you need to squash the 2 commit for have clean history...

AlphaLima: No the first rebase commit out commented out wind  by mistake.
And I re-enabled it with the rebase commit.

brunotl: cf. "squashing commit" : https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History

AlphaLima: And how to do it in Github or Smartgit? I'm a bit afraid of the  git command since I lost some of my commits with it.

brunotl: easy : https://www.google.com/search?q=smartgit+interactive+rebase

AlphaLima:
Does not work with SmartGIT 8

Common/Source/Devices/devLX16xx.cpp
AlphaLima: Does not work with SmartGit8 nor Online.

PR #1310, opened 2019-04-02
Task Overview with additional statistic line and revert button

AlphaLima:
Calc Button and ETE Entry replaced by ETE line
Analyze Button called by pressing Task distance line
Reverse button added
FAI indicator added

image

before:
image


brunotl:
clicking to summary line for Calc is not self document feature, you have place for Calc button...

AlphaLima:
Hmm,  almost every line has such dialog function. e.g. the waypoints in the same dialog.
Maybe not self explaining but intuitive. Give it a try! 
I want to space make bigger buttons, but I can reenable Calc.

PR #1319, opened 2019-04-08
Task Overview improvements (Task Revert, additional ETE line)

AlphaLima:
Task Overview with additional statistic line (ETE) and 
revert button (Fly the task the other way around)
for Task-Calculation (if MC=0, MC = safetyMC)
works for AAT as well

### DIFF
--- a/Common/Data/Dialogs/dlgTaskOverview_L.xml
+++ b/Common/Data/Dialogs/dlgTaskOverview_L.xml
@@ -3,15 +3,17 @@
   <WndForm Name="frmTaskOverview" Type="Dialog" X="0" Y="0" Width="320" Height="240" Caption="_@M688_" Font="2" >
     <WndFrame X="0" Y="0" Width="100" Height="20" Font="2" />
     <WndButton   Name="cmdClose" Caption="_@M186_"    X="2" Y="2" Width="98"  Height="22" Font="2" OnClickNotify="OnCloseClicked"/>
+<!--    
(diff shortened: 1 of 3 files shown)


alus-:
The reverse task button in the task overview is very useful! I was about to start investigating about how to do it but you already did it, nice!

brunotl:
Reverse task are already available in "Nav 3/3" Menu, but you know that you have added it 4 years ago ...

alus-:
Ah, I remember that! The button reverse task briefly appeared in the task overview some year ago...
Anyway it is very handy also for GA: assume that you planned a flight to somewhere, then from there you want to fly back home, it is simply easier and faster to load the task and directly reverse it in the task overview. There is space for the button there so I don't see why no to have it.

brunotl:
agree, but there is no space for button in fact, "Reverse" replace "Analysis", i hope nobody complain because Analysis was removed...

alus-:
Oh sorry I did not realized that the button "Analysis" was replaced, then you are right.
Anyway, with "there is space" I meant that the "Load" button can be made smaller to gain the necessary space to add the "Reverse" button. If I recall carrecly it was done in this way some year ago.
Would be OK this way?

AlphaLima:
We can easily re*enable the analysis button. In fact the ETE Entry was replace by the Reverse button.
However you can reach Analysis by pressing on the Total line!
and Task calculation by pressing on the ETE line!

brunotl:
personaly, i prefert to not remove "Analysis" and "Calc" button if possible because each time we remove something we have complain from users...

AlphaLima:
et voilà

PR #1320, opened 2019-04-09
Better close and verify OpenAIP airspace

alus-:
Tested with airspace .aip from OpenAIP and the test files (with many repeated points) in the ZIP:

[AIPtests.zip](https://github.com/LK8000/LK8000/files/3061125/AIPtests.zip)

### DIFF
--- a/Common/Source/Airspace/LKAirspace.cpp
+++ b/Common/Source/Airspace/LKAirspace.cpp
@@ -1521,26 +1521,24 @@ bool CAirspaceManager::CalculateSector(TCHAR *Text, CPoint2DArray *_geopoints, d
 // Also the geopointlist last element have to be the same as first -> openair doesn't require this, we have to do it here
 // Also delete adjacent duplicated vertexes
 
-void CAirspaceManager::CorrectGeoPoints(CPoint2DArray &points) {
-    if (points.size() == 0) return;
+bool CAirspaceManager::CorrectGeoPoints(CPoint2DArray &points) {
+
+    // Here we expect at least 3 points
+    if(points.size() < MIN_AS_SIZE) return false;
 
     // Close polygon if not closed
-    CPoint2D first = points.front();
-    CPoint2D last = points.back();
-    if ((first.Latitude() != last.Latitude()) || (first.Longitude() != last.Longitude())) points.push_back(first);
+    const CPoint2D first = points.front();
+    if (first != points.back()) points.push_back(first);
 
-    // Delete duplicated vertexes
-    bool firstrun = true;
-    CPoint2DArray::iterator it = points.begin();
-    while (it != points.end()) {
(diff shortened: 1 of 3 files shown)


Common/Source/Airspace/LKAirspace.cpp
```diff
@@ -1521,26 +1521,24 @@ bool CAirspaceManager::CalculateSector(TCHAR *Text, CPoint2DArray *_geopoints, d
...
-        last = *it;
-        ++it;
-        firstrun = false;
+    const CPoint2D& first = points.front();
+    if (first != points.back()) points.push_back(first);
+
+    // Delete duplicated vertexes and equal to first one
+    CPoint2DArray::iterator it = std::next(points.begin()); // Start from second point
+    while(it != std::prev(points.end())) { // until the point before last one
+        if (first == (*it) || (*std::prev(it)) == (*it)) it = points.erase(it);
```
brunotl: why remove all point equal to the first one ?

alus-: This is just an extra check since OpenAIP data demonstrated to not always be fully correct. The first point has to be repeated only at the end to "close" the polygon. So, in theory, when the first point is re-encountered in means: the polygon is now closed and so finished. Let's imagine that the first point is repeated in the middle of the vector, then in practice you will see a line going from there to first point and then back to the next creating 2 polygons which clearly it is not the intention. In theory no other point has to be repeated; anyway to make a nested loop to compare all points with all points it is probably overshooting. Instead while already iterating on the points just check also on the first it is not requiring that much.

brunotl: > creating 2 polygons which clearly it is not the intention

yes or no, we can't be sure, 
even you right, i think trying to solve some source error automaticaly is bad idea  because that hide the pour quality of the data...

still we don't have offical airspace datasource, each pilote need to check if aispaces he see on his instrument are right comparing to the ICAO offical maps...

for illustrate my point of view, taking your exemple file test.aip : 

this is the real airspace shape (maybe) : 
image

this is the shape we have without removing all point equal to first ( we see clearly an error ) : 
image

with removing, it look good, but is not ... : 
image


PS : maybe it's a dream, but if all airspace users become concerned by the aispace data quality and ask to eurocontrol for have AIXM airspaces data, maybe the data will become public one day....

alus-: OK so I pushed another commit where I'm not anymore checking if the first point is repeated, the user will find out looking the map...

brunotl: so, i understand you are agree...

if this case happen lot of time, we can think to add a warning at airspace loading if we detect that ...

alus-: Well, I don't know how often is happening anyway as you pointed out it should be up to OpenAIP to make those checks.

PR #1323, opened 2019-04-11
Shared Port fix

AlphaLima:
better comparison also for TCP/UDP/ CLient /Server and Bluetooth Connections


Common/Source/Comm/device.cpp
```diff
@@ -599,18 +629,21 @@ BOOL devInit() {
             ReadPortSettings(i, Port, &SpeedIndex, &BitIndex);
         }
         // remember: Port1 is the port used by device A, port1 may be Com3 or Com1 etc
+        /*
```
brunotl: useless comment, does not help to understand code ....

AlphaLima: removed

PR #1324, opened 2019-04-11
Task Overview improvements (Task Revert, additional ETE line)

AlphaLima:
calculation also for AAT Tasks
Task Overview with additional statistic line 
revert button added

image


AlphaLima:
squashed version

PR #1326, opened 2019-04-11
Shared Port fix

AlphaLima:
better comparison to find shared ports
shared ports also for TCP, UDP, client/server and Bluetooth connections
correct search for shared ports "behind" master Port

### DIFF
--- a/Common/Source/Comm/device.cpp
+++ b/Common/Source/Comm/device.cpp
@@ -550,6 +550,36 @@ static BOOL devInit(PDeviceDescriptor_t d){
     return(TRUE);
 }
 
+
+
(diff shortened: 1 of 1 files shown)


Common/Source/Comm/device.cpp
```diff
@@ -628,7 +654,6 @@ BOOL devInit() {
                 }
               }
             }
-            continue;
```
brunotl: are you sure ?

AlphaLima: yes, did  not work with it!

AlphaLima: mind, the loop is counting different now

PR #1327, opened 2019-04-11
Shared Port fix

AlphaLima:
skip new devices on shared ports
better comparison also for TCP/UDP Client/Server and Bluetooth Connections

### DIFF
--- a/Common/Source/Comm/device.cpp
+++ b/Common/Source/Comm/device.cpp
@@ -598,7 +598,6 @@ BOOL devInit() {
     pDevPrimaryBaroSource = NULL;
     pDevSecondaryBaroSource = NULL;
 
- //   std::set<tstring> UsedPort; // list of already used port
     
     for (unsigned i = 0; i < NUMDEV; i++) {
         DeviceList[i].InitStruct(i);
--- a/Common/Source/Comm/device.cpp
+++ b/Common/Source/Comm/device.cpp
@@ -550,6 +550,36 @@ static BOOL devInit(PDeviceDescriptor_t d){
     return(TRUE);
 }
 
+
+
+bool IsIdenticalPort(int i, int j)
+{
+  if (_tcscmp(szPort[i], _T("UDPServer")) == 0)
+  {
+    if (dwIpPort[i] == dwIpPort[j])
+      return true;
+  }
+  else
+  if (_tcscmp(szPort[i], _T("TCPServer")) == 0)
+  {
+    if (dwIpPort[i] == dwIpPort[j])
+      return true;
+  }
+  else
+  if (_tcscmp(szPort[i], _T("TCPClient")) == 0)
+  {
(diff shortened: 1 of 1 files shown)


Common/Source/Comm/device.cpp
```diff
@@ -598,45 +627,37 @@ BOOL devInit() {
...
-        // remember: Port1 is the port used by device A, port1 may be Com3 or Com1 etc
+
+	unsigned int j;
+	DeviceList[i].iSharedPort =-1;
+	for( j = 0; j < i ; j++)
+	{
+	  if(!DeviceList[i].Disabled)
+	  if(!DeviceList[j].Disabled)
+	  if( (IsIdenticalPort(i,j)) &&  DeviceList[j].iSharedPort <0)
+	  {
```
brunotl: DeviceList[i].Disabled i always false
no need to call IsIdenticalPort() if DeviceList[j].iSharedPort >= 0

so it's better to write :
```
if( !DeviceList[j].Disabled && (DeviceList[j].iSharedPort < 0) && IsIdenticalPort(i,j) ) {
..
}

```

Common/Source/Comm/device.cpp
```diff
@@ -598,45 +627,37 @@ BOOL devInit() {
...
-          }
-        }
-        UsedPort.insert(Port);
-        
-        // remember: Port1 is the port used by device A, port1 may be Com3 or Com1 etc
+
+	unsigned int j;
+	DeviceList[i].iSharedPort =-1;
+	for( j = 0; j < i ; j++)
+	{
```
brunotl: "j" is local loop iterator so no need to declare them outside

it's better to write 

`for(unsigned j = 0; j < i ; j++)`

Common/Source/Comm/device.cpp
```diff
@@ -550,6 +550,36 @@ static BOOL devInit(PDeviceDescriptor_t d){
...
+  {
+    if (dwIpPort[i] == dwIpPort[j])
+      return true;
+  }
+  else
+  if (_tcscmp(szPort[i], _T("TCPClient")) == 0)
+  {
+    if (_tcscmp(szIpAddress[i], szIpAddress[j]) == 0)
+      if (dwIpPort[i] == dwIpPort[j])
+        return true;
```
brunotl: it's better to test "dwIpPort" before "szIpAddress"

AlphaLima: done:
https://github.com/LK8000/LK8000/pull/1328

Common/Source/Comm/device.cpp
AlphaLima: done:
https://github.com/LK8000/LK8000/pull/1328

Common/Source/Comm/device.cpp
AlphaLima: done:
https://github.com/LK8000/LK8000/pull/1328

PR #1334, opened 2019-04-25
replace LK_tcsncp by  _sntprintf to prevent UTF8 assertion

AlphaLima:
also use correct max string length MAX_LEN (instead of hardcoded 30)


brunotl:
that fix nothing, 
before you have problem because you truncate string to 30 byte, 
after you have same problem because you truncate string to 80 byte...

AlphaLima:
No, I'm 100% sure also
   _sntprintf(Comment, 30 ,_T("%s"), WayPointList[idx].Comment);
fixed that problem and comitted it at first! (try it)
However 30 is not the Comment size (old code artefact), so I squashed the two commits.

brunotl:
if you truncate multibyte encoded string to fixed number of byte you can have incomplete caractere,  no need to test anything, it's fact ...

PR #1336, opened 2019-04-30
Fix FLARM code rename bug then FLARM is feed from LK24

ToninoTarsi:
Now IDFLARM.TXT also work with LT24 radar


brunotl:
merged manually

PR #1337, opened 2019-04-30
DrawHSI: use references...

alus-:
No need to have a "return struct" when we can use references.

### DIFF
--- a/Common/Source/Draw/LKDrawInfoPage.cpp
+++ b/Common/Source/Draw/LKDrawInfoPage.cpp
@@ -982,17 +982,16 @@ label_HSI:
         #else
 	VDrawLine(Surface,rc, qcolumn[0],qrow[2],qcolumn[16],qrow[2],RGB_DARKGREEN);
         #endif
-	static bool showQFU=false;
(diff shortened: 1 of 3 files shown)


Common/Source/Draw/LKDrawInfoPage.cpp
```diff
@@ -982,17 +982,16 @@ void MapWindow::DrawInfoPage(LKSurface& Surface,  const RECT& rc, bool forceinit
         #else
 	VDrawLine(Surface,rc, qcolumn[0],qrow[2],qcolumn[16],qrow[2],RGB_DARKGREEN);
         #endif
-	static bool showQFU=false;
-	static bool showVFRlanding=false;
-	HSIreturnStruct HSI;
-	HSI=DrawHSI(Surface,rc);
-	if(HSI.landing) {
+	static bool showQFU=false, showVFRlanding=false;
+	static bool usingQFU, approach, landing;
```
brunotl: no need static for DrawHsi output parameter...

alus-: You're right! Fixed...

PR #1338, opened 2019-05-02
Parse numbers as radiof requency with decimal only.

AlphaLima:
This prevents airspaces numbers identified as frequency (such as french notam airspace numbering)


brunotl:

PR #1342, opened 2019-05-05
correct handling of unnamed tasks

AlphaLima:
allow comma in waypoint and task turnpoint names


brunotl:
- pull request title as nothing to do with change

- cup file is csv file, before this change, we don't handle field including separator, now we don't handle field containing  escaped double quote, so change a bug to another bug is worth than nothing...

cup file format : http://download.naviter.com/docs/CUP-file-format-description.pdf
csv file format : https://tools.ietf.org/html/rfc4180#page-2

PR #1345, opened 2019-05-07
Task, new value elapsed Task time

AlphaLima:
Task, stop calculation update after valid finish
Task, use old average task speed calculation
Task, new value elapsed Task time


brunotl:
3 change in one commit... that a bad practice,

PR #1368, opened 2019-07-16
critical fix

brunotl:


Common/Source/Oracle.cpp
```diff
@@ -45,8 +45,8 @@ TCHAR *AltDiffToText(double youralt, double wpalt) {
 }
 #endif
 
-TCHAR *WhatTimeIsIt(void) {
-  static TCHAR time_temp[60];
+const TCHAR *WhatTimeIsIt(void) {
+  thread_local static TCHAR time_temp[48];
```
AlphaLima: Does not compile for PNA version!

AlphaLima: #if defined(PNA) && defined(UNDER_CE)
  static TCHAR time_temp[48];
#else
  thread_local static TCHAR time_temp[48];
#endif

brunotl: right fix is to use gcc "__thread" instead of c++11 "thread_local" for gcc older than 4.8

Issue #1119, opened 2017-06-16
Move waypoints in Task Edit (PG optimized or AAT task)


parapenT1sta:
Situation 1 with these Task Settings in System Setup: 
![1](https://user-images.githubusercontent.com/608650/27225089-8f5e546a-5290-11e7-9802-a0bf635d8777.JPG)

Create a task with 3 waypoints, if I move the middle waypoint up to start position or down to finish positon and it doesn't move the radius info, only the waypoint. Is this normal? I thought it should move all the info that we setup while introducing the waypoint in the task.

Situation 2 with these Task Setings in the System Setup:
![2](https://user-images.githubusercontent.com/608650/27225241-31f5b04c-5291-11e7-9068-f25d2ca0f0a0.JPG)

Create a task with 3 waypoints. 
Start: Cylinder, 2Km
Turnpoint: Cylinder, 1Km
Finish: Cylinder, 0.4km

When I move up the finish waypoint to the middle position, it converts the waypoint from Cylinder to Circle ESS even when Sector is defined as "Line" in the system setup.

brunotl:
need refator Start and Finish radius storing for AAT and PG optimized task

parapenT1sta:
@brunotl Does it make sense having Task settings in the system setup because every task is different every day, at least for us paragliding pilots.

What is your opinion about it?

brunotl:
yes it's make sens, because it's unique setting value for default glider task, and it's default value for all other ( AAT or PG Optimized )

pventafridda:
same for gliders. but then every task parameter must be exposed in more 
than one page in task setup.



In data 21 Giugno 2017 4:27:53 PM Sérgio Da Silva 
<notifications@github.com> ha scritto:

AlphaLima:
https://github.com/LK8000/LK8000/issues/1357

parapenT1sta:
@AlphaLima The version you sent me doesn't fix the second situation that I mention here. I am using Paragliding Mode. It doesn't happen all the times and It happens more when I have Optimize Route On.

Create a task with 3 waypoints.
Start: Cylinder, 2Km
Turnpoint: Cylinder, 1Km
Finish: Cylinder, 0.4km

When I move up the finish waypoint to the middle position, it converts the waypoint from Cylinder to Circle ESS even when Sector is defined as "Line" in the system setup.

parapenT1sta:
[...]

PR #1375, opened 2019-07-28
Fix waypoint Move glitch:

AlphaLima:
https://github.com/LK8000/LK8000/issues/1119

if AAT
copy entire task point settings by memcpy 
else
exchange waypoint index only


brunotl:
are you sure that fix #1119 ?

parapenT1sta:
@AlphaLima If you send me the file with this merge I can test it. On that issue 1119 I have two situations, It might fix one and not the second one.

PR #1397, opened 2019-12-31
refactoring localized language files

brunotl:
use "flat" json for compatibility with translation tools.

using our own format for language files forbid to use online translation platform like weblate.org, first idea could be to write tools to convert our files to format supported by this platform and back ... but that add 2 new source of bugs.

using json instead of our own format have few advantage :
  - no need to convert files to use translation tools
  - files can be read using well tested json parser library


brunotl:

PR #1405, opened 2020-02-10
There is no "NOTAM" airspace in openAIP

alus-:
As far as I know there is no "Notam" airspace category in _openAIP_. (There is in _OpenAir_: `AC NOTAM`).
For reference here the _openAIP_ format specifications: [Airspace.pdf](https://github.com/LK8000/LK8000/files/4182998/Airspace.pdf)

### DIFF
--- a/Common/Source/Airspace/LKAirspace.cpp
+++ b/Common/Source/Airspace/LKAirspace.cpp
@@ -2189,9 +2189,6 @@ bool CAirspaceManager::FillAirspacesFromOpenAIP(const TCHAR* szFile) {
               if (_tcsicmp(dataStr,_T("GLIDING"))==0) Type=GLIDERSECT;
             }
             break;
-        case 'N':
-          /*  if (_tcsicmp(dataStr,_T("NOTAM"))==0) */ Type=CLASSNOTAM; // Prohibited area
-            break;
         //case 'O':
             //if (_tcsicmp(dataStr,_T("OTH"))==0) continue; //TODO: OTH missing in LK8000
             //break;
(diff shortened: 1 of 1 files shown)


alus-:
Correct _openAIP_ does not support NOTAM's as airspace definition. 
The current airspace definitions supported by _openAIP_ are:
- `A`
- `B`
- `C`
- `CTR`
- `D`
- `DANGER`
- `E`
- `F`
- `G`
- `GLIDING`
- `OTH`
- `RESTRICTED`
- `TMA`
- `TMZ`
- `WAVE`
- `PROHIBITED`
- `FIR`
- `UIR`
- `RMZ`

Why should it be `N` of course and not `NOTAM` or something else?
Was the intent of this to predict what _openAIP_ will do?
As far as I know the next version of _openAIP_ will be Json based and nobody except _openAIP_ is generating aip files. And then why anybody would be willing to list point by point in _openAIP_ while in _OpenAir_ is possible to make use of arcs an circumferences?
In the current _openAIP_ format it will never happen: `<ASP CATEGORY="N">` therefore it makes no sense to expect for that.

AlphaLima:
I don't see any disadvantage, but I don't mind, I'm tired about arguing about NOTAMS.
https://www.postfrontal.com/forum/topic.asp?TOPIC_ID=9027

alus-:
The discussion is about _OpenAir_. The categories in _OpenAir_ are identified at line 1741 comparing against `k_strAreaStart` string vector. While in _openAIP_ the categories are only as from their format specification.

PR #1420, opened 2020-03-21
Added Support for FANET-Device

gereic:
Weather-Data will be displayed on Map with windsock, windspeed and windgust.


gereic:
Added support for diplaying FANET-Messages

brunotl:
merged manually

PR #1421, opened 2020-03-23
Acknloage Airspace Pictorial

AlphaLima:
draw pictorial outline if airspace is acknoledged

### DIFF
--- a/Common/Header/LKAirspace.h
+++ b/Common/Header/LKAirspace.h
@@ -131,7 +131,7 @@ public:
   // Set ack validity timeout according to config prameter
   void SetAckTimeout();
   // check if airspace was already acknolaged
-  bool Acknowledged(void);
+  const bool Acknowledged(void) const;
   // get nearest distance info to this airspace, returns true if distances calculated by warning system
   bool GetDistanceInfo(bool &inside, int &hDistance, int &Bearing, int &vDistance) const;
   // get warning point coordinates, returns true if airspace has valid distances calculated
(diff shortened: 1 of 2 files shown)


Common/Header/LKAirspace.h
```diff
@@ -131,7 +131,7 @@ class CAirspaceBase
   // Set ack validity timeout according to config prameter
   void SetAckTimeout();
   // check if airspace was already acknolaged
-  bool Acknowledged(void);
+  const bool Acknowledged(void) const;
```
brunotl: const is useless for return by value

brunotl: const ise useless on return by value.

Common/Source/Airspace/LKAirspace.cpp
```diff
@@ -624,7 +624,7 @@ void CAirspaceBase::SetAckTimeout() {
     _warnacktimeout = _now + AcknowledgementTime;
 }
 
-bool  CAirspaceBase::Acknowledged(void) {
+const bool  CAirspaceBase::Acknowledged(void) const {
```
brunotl: same as declaration

Common/Source/Airspace/LKAirspace.cpp
```diff
@@ -3603,8 +3603,11 @@ void CAirspace::DrawPicto(LKSurface& Surface, const RECT &rc) const {
         const auto oldColor = Surface.SetTextColor(TypeColor());
 
         const auto oldPen = Surface.SelectObject(FramePen);
+bool Fill = Enabled();
+if(Acknowledged())
+   Fill = false;
```
brunotl: invalid code indent

PR #1425, opened 2020-04-06
NettoVario

AlphaLima:
Don't set NettoVario to zero before takeoff in non SIM mode.
https://www.postfrontal.com/forum/topic.asp?whichpage=2&TOPIC_ID=9061&#71924


brunotl:
Why change is needed ? 

before takeoff we are outside of valid part of polar, so we can't calculate Netto, so ,correct fix should be to display "---" (no value) instead of "0.0".

brunotl:
this revert previous change : https://github.com/LK8000/LK8000/pull/1201

PR #1426, opened 2020-04-09
Auto Radio B> Option

AlphaLima:
Option to automatically set the best alternate frequency. Separate for active and Passive Radio frequency

image

### DIFF
--- a/Common/Source/Dialogs/dlgRadioSettings.cpp
+++ b/Common/Source/Dialogs/dlgRadioSettings.cpp
@@ -291,6 +291,22 @@ TCHAR Name[250];
           wpnewVol->SetCaption(Name);
         }
     }
+
+    WindowControl* wAuto = wf->FindByName(TEXT("cmdAutoActive"));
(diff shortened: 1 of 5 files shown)


Common/Source/Dialogs/dlgRadioSettings.cpp
```diff
@@ -239,20 +239,20 @@ TCHAR Name[250];
     {
         if( ActiveRadioIndex > RESWP_END)
         {
-          _stprintf(Name,_T("%s"),WayPointList[ActiveRadioIndex].Name);
+          _sntprintf(Name,DEVICE_NAME_LEN,_T("%s"),WayPointList[ActiveRadioIndex].Name);
```
brunotl, line 242: Name size is 250 and WAYPOINT::Name is 31 so :
 - previous _stprintf usage can't overflow  -> no bug
 - using _sntprinf with size less than 31 can truncate string -> bug on utf8 platform

PR #1428, opened 2020-04-09
Radio auto select

AlphaLima:
reverted _snprintf


brunotl:
last commit don't work on unicode TARGET ( WinCE, Win32 ... ) and we don't need another function to truncate string ...

PR #1434, opened 2020-04-21
refactoring char* to/from wchar_t*  string copy

brunotl:
add few function and function helper to cover all use case :

``` cpp
size_t to_utf8(const TCHAR* string, char* utf, size_t size);

template<size_t size> 
size_t to_utf8(const TCHAR* string, char (&utf8)[size]);

size_t from_utf8(const char* utf, TCHAR* string, size_t size);

template<size_t size>
size_t from_utf8(const char* utf8, TCHAR (&string)[size]);

size_t to_usascii(const char* utf8, char* ascii, size_t size);
size_t to_usascii(const wchar_t* unicode, char* ascii, size_t size);

template<typename CharT, size_t size>
size_t to_usascii(const CharT* source, char (&ascii)[size]);
```
all of them return length of output string and check for buffer overflow.

this function is made to replace the 10 old platform specific or bugged `TCHAR2xxx()` function


AlphaLima:
works for me!

PR #1441, opened 2020-04-29
show horizontal airspace distance also if in/above/under the airspace

AlphaLima:
show values colored if inside.


brunotl:
you can't use `LKFormatValue` return value to change color, that also hide unit ...

AlphaLima:
your right right, was just an idea.
maybe we can format the strings "above" below "below",  but I'm afraid we cant extract this information from the vertical distance.

brunotl:
Merged Manualy

PR #1442, opened 2020-04-29
Nano3 igc fix

AlphaLima:
After some long test sessions with walter per TeamViewer we checked now all functions on a S100 as well, but also needed the old TCHAR2usascii function.
All worked fine then.
-setup NMEA sentences
-Task declaration
-IGC File download (on IGC device, no crash on none IGC)
-bi-directional exchange of MC, BAL, BUGS, STF, ...
-Setup and Value Dialog as part of the Device config dialog


Common/Data/Language/Translations/en.json
```diff
@@ -2539,5 +2539,19 @@
     "_@M001911_": "uartPassthrough",
     "_@H001343_": "Check to pass data received by U2 into U1",
     "_@M001912_": "uart1Raw",
-    "_@H001344_": "Check to make U1 data transferred raw instead of line by line"
```
brunotl, line 2542: _@H001344_ is used by blueflyvario, you can't remove them ...

PR #1445, opened 2020-05-05
Fix device dialog mixup due to disabled TESTBENCH switch

AlphaLima:


Common/Source/Dialogs/dlgConfiguration.cpp
```diff
@@ -1745,23 +1744,24 @@ static void setVariables( WndForm *pOwner) {
         {
           if (_tcscmp(devRegisterGetName(i), deviceName1) == 0)
           {
-            StartupStore(_T("==================================================== %s"),NEWLINE); // 091105
+
             dwDeviceIndex[devIdx] = i;
-            StartupStore(_T("........... SetVariable Device #%i id:%i %s %s"),devIdx,  dwDeviceIndex[devIdx], deviceName1,NEWLINE); // 091105
             ReadPortSettings(devIdx,szPort,NULL, NULL);
+#ifdef TESTBENCH
```
brunotl, line 1750: ReadPortSetting should be inside "#idef TESTBENCH"

PR #1450, opened 2020-05-10
Add device connection type NMEA Replay (device emulation)

AlphaLima:
NMEA replay setup in device dialog
exclude NMEA replay from shared ports.
Activate Start/Stop Buttons in File dlg (= COM Port restart)
add new Port type FilePort for NMEA File input for each device seperatly
Allow parallel NMEA reaply


brunotl (changes requested):
after quick test : 
  - replay rate meaning is not clear ( if i select x1 it seem's 1 second duration is around 3 second )
  - replay 2 or more file with different replay rate is non sens ...

 
looking the code, i see lot of mixing/wrong code indent, we have already too much of that in code, not need to add more ...

PR #1453, opened 2020-05-24
Enter own FLARM ID in Flarm Dialog

AlphaLima:
Ignore own FLARM ID Ghost in case of mutliple FLARM Sources (e.g. OGN).


brunotl:
dangerous feature if used by pilote using club gliders ....

instead we can request FLARM_ID to connected device ( cf. $PFLAC command )

AlphaLima:
Two other aspects:
-not every Flarm is bi-directional
-some use a virtual Flarm/OGN address (Glidertracker hub)
-for different club gliders, we store it in the devices config that need to be changed in that case anyway.

brunotl:
`-for different club gliders, we store it in the devices config that need to be changed in that case anyway.

that is the problem, if you don't change it you "hide" another glider, and this glider can fly close to you... Flarm is made to prevent collision and this feature must be thinking with security point of view first.

PR #1459, opened 2020-06-05
fix missing last waypoint in CUP Task

AlphaLima:
-simplified CUP task load
-remove doubled waypoints
-prefer airfields over waypoint if double defined (GA)
according to this bug report:
https://www.postfrontal.com/forum/topic.asp?whichpage=1&TOPIC_ID=9115&#72126

### DIFF
--- a/Common/Source/SaveLoadTask/LoadCupTask.cpp
+++ b/Common/Source/SaveLoadTask/LoadCupTask.cpp
@@ -257,104 +257,51 @@ public:
 bool ConvertStringToTask( LPCTSTR szTaskSteing,   mapCode2Waypoint_t &mapWaypoint)
 {
   std::vector<tstring>  Entries =   CupStringToFieldArray(szTaskSteing);
-  bool bTakeOff = true;
-  bool bLoadComplet = true;
+
 
   size_t Idx =0;
   size_t idxTP = 0;
-  TCHAR TpCode[NAME_SIZE + 1];
 
-  WAYPOINT* WPtoAdd=NULL;
 
-  while (bLoadComplet && (Idx < (Entries.size()))) {
+    for (uint i =0; i < Entries.size() ; i++)
+    {
+      mapCode2Waypoint_t::iterator It = mapWaypoint.find(Entries[Idx++].c_str());
+      if(It != mapWaypoint.end())
+      {
+        int ix = FindOrAddWaypoint(&(It->second),true);  // first try to find airfield
+        if(ix < 0)
+          ix= FindOrAddWaypoint(&(It->second),true); // not found try if waypoint exist
 
-      if (idxTP < MAXTASKPOINTS) {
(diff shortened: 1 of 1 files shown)


Common/Source/SaveLoadTask/LoadCupTask.cpp
```diff
@@ -257,104 +257,51 @@ class CupObsZoneUpdater {
...
+
+    for (uint i =0; i < Entries.size() ; i++)
+    {
+      mapCode2Waypoint_t::iterator It = mapWaypoint.find(Entries[Idx++].c_str());
+      if(It != mapWaypoint.end())
+      {
+        int ix = FindOrAddWaypoint(&(It->second),true);  // first try to find airfield
+        if(ix < 0)
+          ix= FindOrAddWaypoint(&(It->second),true); // not found try if waypoint exist
+
```
brunotl, line 274: line 271 and 273 are indentical so 272 and 273 can be removed

Common/Source/SaveLoadTask/LoadCupTask.cpp
```diff
@@ -257,104 +257,51 @@ class CupObsZoneUpdater {
...
+#ifdef DEBUG_TASK_LOAD
+#include "Globals.h"
+  int i =0;
+  while ( Task[i].Index  >=0)
+  {
+    StartupStore(_T(".......Taskpoint: %u %s\n"),(unsigned)(i), WayPointList[Task[i].Index].Name );
+    i++;
   }
-   return bLoadComplet;
+#endif
```
brunotl, line 297: line 289 to 298 are useless, we have debugger ...
worst, including Global.h inside function body is big mistake, it's undefined behavior ...

Common/Source/SaveLoadTask/LoadCupTask.cpp
```diff
@@ -257,104 +257,51 @@ class CupObsZoneUpdater {
...
+
+
+if(idxTP > 0)
+  return true;
+else
+  return false;
```
brunotl, line 303: could be simply :
```cpp
return (idxTP > 0);
```

Common/Source/SaveLoadTask/LoadCupTask.cpp
AlphaLima: yes, was a mistake commig from copying already fixed in a new commit.

Common/Source/SaveLoadTask/LoadCupTask.cpp
AlphaLima: way already disabled, with the second commit

Common/Source/SaveLoadTask/LoadCupTask.cpp
AlphaLima: done

AlphaLima:
squashed and pushed with another commit

PR #1471, opened 2020-07-03
Prevent drawing airspace on map

AlphaLima:
in case it is defined on MSL (=on/under ground) without any vertical expansion.
For example (in open air format):
AH 0

This can be useful to have regional radio frequencies present.

https://www.postfrontal.com/forum/topic.asp?whichpage=1&TOPIC_ID=9124&#72205


brunotl:
only when "Fill Border Only" drawing is selected ?

PR #1473, opened 2020-07-08
Nano3 Get Polar from LX device

AlphaLima:
optionally read Polar coefficients and name from external LX device, instead of own polar.


brunotl:
fixed and merged manualy

PR #1475, opened 2020-07-09
No Takeoff and landing for Task declaration by default.

AlphaLima:
If realy needed it can be re-enabled by
DeclareTakeoffLanding=1
in the Profile file.

### DIFF
--- a/Common/Source/Devices/devVolkslogger.cpp
+++ b/Common/Source/Devices/devVolkslogger.cpp
@@ -123,7 +123,7 @@ BOOL VLDeclare(PDeviceDescriptor_t d, Declaration_t *decl, unsigned errBufferLen
   TCHAR2usascii(decl->CompetitionClass, vl.declaration.flightinfo.competitionclass, 13);
   vl.declaration.flightinfo.competitionclass[12]='\0'; // BUGFIX 100331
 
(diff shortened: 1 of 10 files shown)


Common/Source/Devices/devVolkslogger.cpp
```diff
@@ -123,7 +123,7 @@ BOOL VLDeclare(PDeviceDescriptor_t d, Declaration_t *decl, unsigned errBufferLen
   TCHAR2usascii(decl->CompetitionClass, vl.declaration.flightinfo.competitionclass, 13);
   vl.declaration.flightinfo.competitionclass[12]='\0'; // BUGFIX 100331
 
-  if (ValidWayPoint(HomeWaypoint)) {
+  if (ValidWayPoint(HomeWaypoint) && DeclTakeoffLanding) {
```
brunotl, line 126: are you sure ?

"vl.declaration.flightinfo.homepoint.name" is not part of turnpoint list.

AlphaLima: No I'm not sure but,  but what is this declaration about? What is it good for?
It is also skipped if no valid homewaypoint is defined.

AlphaLima: changed back

PR #1479, opened 2020-07-25
Eink android

tvebi:
https://www.postfrontal.com/forum/topic.asp?TOPIC_ID=9125

due to a hint from "brunotl" I have compiled some classes to set the EPD modes on my alfapilot device (boyue T62 clone).
In a stand alone application  I had success setting this modes by EPDAction. So I think in principal this will work.

I tried the same in LK8000    "NativeView" but had no success, may be the C++ part does not render to NativeView.
I am not familiar with that back and forth from C++ and Java so I am a little bit lost to get the right View to set the EPDmode.

So if anyone has an idea let me know


brunotl:
i think it should work if you put at the end of LK8000.initSDL() method something like this :
`    EPDFactory.getEPDController().setEpdMode(nativeView, 0, 0, 0, 0, 0, 0, "EPD_A2");`

tvebi:
> 
> 
> i think it should work if you put at the end of LK8000.initSDL() method something like this :
> ` EPDFactory.getEPDController().setEpdMode(nativeView, 0, 0, 0, 0, 0, 0, "EPD_A2");`


at this point in LK8000.initSDL()  I got  "too early to call this, it has to got a parent yet" (tried this earlier) so

I called it in NativeView.run before runNative();

EPDFactory.getEPDController().setEpdMode(this, 0, 0, 0, 0, 0, 0, "EPD_A2");

This worked but has no effect on on LK8000 rendering.

PR #1480, opened 2020-07-26
compiler error fix

AlphaLima:
changed logical to bitwise AND ( compiler error fix)
always use APT buffer as target from LK destination.
allow STF switch bi-directional exchange

### DIFF
--- a/Common/Source/Devices/devLX_EOS_ERA.cpp
+++ b/Common/Source/Devices/devLX_EOS_ERA.cpp
@@ -1988,7 +1988,7 @@ TCHAR szName[MAX_VAL_STR_LEN];
     if((WayPointList[overindex].Flags & LANDPOINT)> 0) 
     {
       landable = 1;
-      rwdir    = (WayPointList[overindex].RunwayDir;
+      rwdir    = WayPointList[overindex].RunwayDir;
     }
 
     _sntprintf( szTmp,MAX_NMEA_LEN, TEXT("LXDT,SET,NAVIGATE,%i,%s,%i,%i,%i,%i,%s,%i"),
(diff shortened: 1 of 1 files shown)


Common/Source/Devices/devLX_EOS_ERA.cpp
```diff
@@ -1983,15 +1984,11 @@ TCHAR szName[MAX_VAL_STR_LEN];
   {                                  
     int rwdir = 0;
```
brunotl, line 1985: always "0" ...

Common/Source/Devices/devLX_EOS_ERA.cpp
```diff
@@ -405,6 +405,7 @@ int PortNum = d->PortNumber;
     dfe->addEnumText(MsgToken(491)); // LKTOKEN  _@M491_ "OFF"
     dfe->addEnumText(MsgToken(2452)); // LKTOKEN  _@M2452_ "IN"
     dfe->addEnumText(MsgToken(2453)); // LKTOKEN  _@M2453_ "OUT"
+    dfe->addEnumText(MsgToken(2454)); // LKTOKEN  _@M2454_ "IN & OUT"
```
brunotl, line 408: unrelated to change in "DevLX_EOS_ERA::PutTarget"

AlphaLima: right! should be related to STF switch!

AlphaLima: that is correct as it is it is intendend to have the possibility to trigger STF from the device and LK.

Common/Source/Devices/devLX_EOS_ERA.cpp
AlphaLima: ... that is correct in case it is no landaable
changed it in next commit for lanables

PR #1485, opened 2020-08-05
EOS/ERA/10k fixes & IGC File download

AlphaLima:
EOS/ERA/10k fixes & IGC File download
-fix altitude and airspeed trnasormation becuse LX Navigation devices already give true airspeed and true altitude
-IGC file download using binary protocol
-EOS/ERA/10k need at least Firmware revision 1.6x otherwise not fully functional
-fixed QNH direction selection
Important:
-also added QNH direction to Nano3 driver (else critical bug due to shred dialog resource)


brunotl:
- what is ICH ?
- why lx nano3 change have "LX EOS.." title ?
- why airspeed/qnh and igc file download change are in same commit ?
- firmeware requirerement should be comment of commit 9ff51e4...
- PR comment don't explain what is in the PR but only a half of change made in 1 commit...

PR #1490, opened 2020-08-19
Eos IGC download speedup /Radio detect

AlphaLima:
EOS/ERA/LX10k IGC download speedup
auto detect if radio is connected to EOS/ERA/LX10k  (by parsing radio error message)


AlphaLima:
wait

PR #1491, opened 2020-08-26
EOS/ERA/LX10k driver improvements

AlphaLima:
-detect if radio is connected and disable radio communication if not connected
-parse sensor data from broadcast commands
-speedup IGC download


brunotl:
- what is "bradcast" ?
- what do you call "muti instances" ?
- when ambigous resource lock call can happen ?

PR #1493, opened 2020-08-28
check for uninitialized page number during startup

AlphaLima:
prevent assertion error in GA mode


brunotl:
wrong fix, assert is made to detect function call with invalid parameter, so this invalid call must be fixed ...

AlphaLima:
this line can be removed, parameter check in the following line anyway.

AlphaLima (commented):
assert can be removed, parameter is checked anyway

PR #1497, opened 2020-10-05
hide airspace 0m MSL top

AlphaLima:
don't draw airspaces defined with 0m MSL top also for pattern and outline modes as well like semi transparent.
This is used to hide specific airspaces e.g. for radio frequencies of FIS (Flight information service) regions.


brunotl:
- same expression exist 5 time, must be a `CAirspaceBase` method including "`DrawStyle() != adsHidden`" to simplify condition  
  this method could be named `IsVisible()`
image

AlphaLima:
OK, done!

brunotl:
already done

PR #1499, opened 2020-11-15
store serial numbers in device descriptor (not in  global GPS data)

AlphaLima:
Important if two devices are connected, otherwise the DoStatusMessage is toggeling between the devices

fixed vario source in status dialog, same problem

### DIFF
--- a/Common/Header/device.h
+++ b/Common/Header/device.h
@@ -167,7 +167,11 @@ struct DeviceDescriptor_t {
   unsigned ErrTx;
   // Com ports hearth beats, based on LKHearthBeats
   unsigned HB;
-  
+#ifdef DEVICE_SERIAL
+  int HardwareId;
+  int SerialNumber;
+  double SoftwareVer;
+#endif
   NMEAParser nmeaParser;
 //  DeviceIO PortIO[NUMDEV];
   void InitStruct(int i);
--- a/Common/Source/Calc/Vario.cpp
+++ b/Common/Source/Calc/Vario.cpp
@@ -19,7 +19,8 @@ bool VarioAvailable(const NMEA_INFO& Info) {
 }
 
 void UpdateVarioSource( NMEA_INFO& Info, const DeviceDescriptor_t& d, double Vario) {
-  if((unsigned)d.PortNumber <= Info.VarioSourceIdx) {
+  if((unsigned)d.PortNumber <= Info.VarioSourceIdx) 
+  {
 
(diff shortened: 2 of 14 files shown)


Common/Source/Calc/Vario.cpp
```diff
@@ -19,7 +19,8 @@ bool VarioAvailable(const NMEA_INFO& Info) {
 }
 
 void UpdateVarioSource( NMEA_INFO& Info, const DeviceDescriptor_t& d, double Vario) {
-  if((unsigned)d.PortNumber <= Info.VarioSourceIdx) {
+  if((unsigned)d.PortNumber <= Info.VarioSourceIdx) 
+  {
```
brunotl, line 23: why ?

Common/Source/Dialogs/dlgStatus.cpp
```diff
@@ -239,22 +239,17 @@ static void UpdateValuesSystem() {
     if (VarioAvailable(GPS_INFO)) {
 	// LKTOKEN  _@M199_ = "Connected"
 #ifdef DEVICE_SERIAL
-     if(GPS_INFO.HardwareId >0)
+     if(DeviceList[ GPS_INFO.VarioSourceIdx ].HardwareId >0)
```
brunotl, line 242: GPS_INFO and DeviceList is shared data, mutex lock is required.

Common/Header/device.h
```diff
@@ -167,7 +167,11 @@ struct DeviceDescriptor_t {
   unsigned ErrTx;
   // Com ports hearth beats, based on LKHearthBeats
   unsigned HB;
-  
+#ifdef DEVICE_SERIAL
+  int HardwareId;
+  int SerialNumber;
+  double SoftwareVer;
+#endif
```
brunotl, line 174: why store number instead of original string ?

Common/Source/Calc/Vario.cpp
AlphaLima: forgot to restore! can stay original

Common/Source/Dialogs/dlgStatus.cpp
AlphaLima: How, correctly?

Common/Header/device.h
AlphaLima: Just moved it from the without changing data format.

brunotl:
don't build if "DEVICE_SERIAL" is not defined

PR #1504, opened 2020-12-10
No checksum if sending '$$$' from the terminal on KOBO

ToninoTarsi:
The command '$$$' put a RN43 BT module in command mode. No checksum is required in that case.

TODO. Maybe an option to enable/disable/auto the checksum will be more flexible.


brunotl:
ifdef KOBO is useless, you can remove it

brunotl:
merged manually

brunotl:
merged

PR #1509, opened 2020-12-15
support of Silicon Labs CP210x USB to UART bridge controller on LK8000 for Android

lyusupov:


brunotl:
should be also  added to "android/res/xml/usb_device_filter.xml"  to allow auto launch app when device is connected.

lyusupov:
Thank you for your guidance!
I've updated "android/res/xml/usb_device_filter.xml" as well.

PR #1511, opened 2020-12-16
mark Radio Parameter as changed 

AlphaLima:
mark Radio Parameter as changed when changed outside the radio dialog
This will update the radio dialog when changed outside  (e.g. by Best Alternate).


brunotl:
thet made loop : 
 - `devPutFreqActive` set `Changed` true
 - if Changed is true `OnRemoteUpdate` call `devPutFreqActive`

PR #1520, opened 2020-12-23
fix MapScale Offset for 100% TopView in Multimaps

AlphaLima:
we need an own BottomSize corrected RECT for that
otherwise we loose the sideview scaling

image


brunotl:
- don't build on Win32 platform
- why to add BottomSize to Top and Bottom of rc ?
- your fix is workaround for a wrong usage DrawMapScale rc parameter, it's ok, but require comments to explain why we need to add BootomSize to rc...

PR #1521, opened 2020-12-23
added Home distance in NM

AlphaLima:
added
Home required Efficency
Home BRG
Home Dist NM
as requested:
https://www.postfrontal.com/forum/topic.asp?TOPIC_ID=9180


brunotl:
why to add Home Distance in NM ?

right things should be "Home distance in Alternate Unit" ( km if you use NM or NM if you use NM )....

brunotl:
manually merged

PR #1524, opened 2021-01-02
Flarm read fix

AlphaLima:


brunotl:
how replacing 'if(ThreadState)' by 'switch(ThreadState)' can take care only one state processed on one call ?

PR #1534, opened 2021-01-31
Fix display of target distance in upper right corner

AlphaLima:
was drawn under north arrow (when clock disabled)!


Common/Source/Draw/LKDrawLook8000.cpp
```diff
@@ -289,7 +289,7 @@ void MapWindow::DrawLook8000(LKSurface& Surface, const RECT& rc) {
         if ( !OverlayClock && ScreenLandscape && (!((gTaskType==TSK_GP) && UseGates()))) {
             _stprintf(BufferValue + _tcslen(BufferValue), _T(" %s"), BufferUnit);
              Surface.GetTextSize(BufferValue, &TextSize);
```
brunotl, line 291: seem's this line is no more needed...

AlphaLima: right,
 done!

PR #1537, opened 2021-03-15
allow finding 0MSL Top airspaces (not drawn) in Multiselect dialog

AlphaLima:
https://www.postfrontal.com/forum/topic.asp?whichpage=1.66666666666667&TOPIC_ID=9124&#72595


brunotl:
Visible in Multiselect unconditionally ? even if class disabled or hidden by the vertical filter configured ?

AlphaLima:
Yes, that was the idea, because it is implemented for area related radio frequencies only.
So it does not make sense to disable it, vertical filters will curious the feature.

PR #1548, opened 2021-07-12
changed icon of Flarm-object, which are on ground

gereic:
(speed == 0) to black circle

### DIFF
--- a/Common/Source/Draw/DrawFlarmRadar.cpp
+++ b/Common/Source/Draw/DrawFlarmRadar.cpp
@@ -60,26 +60,26 @@ static int nEntrys=0;
 
 typedef struct
 {
-	double fx;
-	double fy;
-	double fAlt;
-	double fFlarmBearing;
-	int iColorIdx;
-	TCHAR szGliderType[FLARMID_SIZE_NAME+1];
+  double fx;
+  double fy;
+  double fAlt;
+  double fFlarmBearing;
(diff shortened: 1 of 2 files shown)


gereic:
2 Loops, because of the Z-Level of the objects.
So static objects are always in Background und the flying ones are on Top.

brunotl:
> 2 Loops, because of the Z-Level of the objects.
> So static objects are always in Background und the flying ones are on Top.

it's not obvious reading the code, so comments is required to avoid mistake in future.

a better approch is to sort object by priority using distance / Moving / Static and draw the N first ...

gereic:
Changed the intend,
create more comment.
In order to sort the list, I think it would be good to create (or use the FlarmId.txt) as a Buddy(Friends)-List.
So that the Buddys will prior shown.
Then we can create another choice for showing flarm-traffic on map
- OFF
- only Friends
- All

What do you think about it ?
When this is done, the sorting make sense I think.

brunotl:
2 point : 
- changing the wall code indent and code flow in same commit make code review impossible
- Flarm is made to prevent collision not to do a buddies tracking even if flarm object is used to draw both on map, so filtering object is not a good idea.

gereic:
So what do I have to do now ??

gereic:
How can we go further?
Any news?

brunotl:
merged manually...

Issue #1550, opened 2021-08-02
Change maximum waypoint radius admited on LK8000 from 100Km to as is possile


Mrapari:
During last Italian campionship I've seen that the maximum waypoint radius admited on LK8000 is 100Km, and for this reason I couldn't use LK because some waypoints radius on task competition was upper 100Km. I think it is easy to solve and I'll be very gratefull if it will done. Thanks so much

AlphaLima:
what version are you you using?
with the latest beta 7.3.8, I don't have any radius limitation, not in normal sectors nor cylindwers.
Testet for normal and AAT Tasks.

![image](https://user-images.githubusercontent.com/1188401/127813875-67e81da6-8d89-4d01-89a4-753a2dc73bcc.png)

Mrapari:
Ho, the problem is shown only for the start point. The other seems to be goodIl 2 Ago 2021 08:29, AlphaLima ***@***.***> ha scritto:
what version are you you using?
with the latest beta 7.3.8, I don't have any radius limitation, not in normal sectors nor cylindwers.
Testet for normal and AAT Tasks.

AlphaLima:
A start radius > 100km ???
(Does not make sense to me)

Mrapari:
I don't know the sense, but it was in the last Italian champions. And for that reason I can't use LK8000 for that tasksIl 2 Ago 2021 09:57, AlphaLima ***@***.***> ha scritto:
A start radius > 100km ???
(Does not make sense to me)

brunotl:
> I don't know the sense, but it was in the last Italian champions. And for that reason I can't use LK8000 for that tasks

can you share the task details ?

Mrapari:
Here one of the task planned on xctrack..Il 2 Ago 2021 11:29, Bruno de Lacheisserie ***@***.***> ha scritto:

I don't know the sense, but it was in the last Italian champions. And for that reason I can't use LK8000 for that tasks

can you share the task details ?

brunotl:
> Here one of the task planned on xctrack..

where is the task ?

Mrapari:
File attached in the previous mailIl 2 Ago 2021 13:42, Bruno de Lacheisserie ***@***.***> ha scritto:

Here one of the task planned on xctrack..

where is the task ?

brunotl:
no attachement in github issues comments ...

Mrapari:
Sorry I've attached in the mail. but now added also on this post
[...]

PR #1551, opened 2021-08-02
allow task start radius up to 1000 distance units (not 100 as before)

AlphaLima:
https://github.com/LK8000/LK8000/issues/1550


brunotl:
only for landscape mode ?
start radius parameter also exist in system config…

PR #1563, opened 2022-01-12
Stop receive thread for task declaration

AlphaLima:
Stop receive thread for task declaration for propper Bluetooth declarations
Fix unintended pilot name  modification due to multi driver instances


brunotl (changes requested):
2 bug fix require 2 commit ...

in all case the 2 fix made in this commit are incomplete or add a new bug...

 - first bug : missing StopRxThread
if stop failed, return is made without calling Declare(false)
     
 - second bug : side effect on "lkDecl" parameter
bug is the "devDeclare" function declaration.
"devDeclare" second parameter is input and must be const...

AlphaLima:
replaced by this request:
https://github.com/LK8000/LK8000/pull/1564

PR #1567, opened 2022-01-20
500ms timeout is to optimistic on some devices.

AlphaLima:
some devices to have a lowercase other uppercae hex checksum => reduced expect check to *5 checksum (5c  would not find 5C)


brunotl:
what is the relation between  "500ms timeout" and "checksum case" ?

AlphaLima:
> what is the relation between "500ms timeout" and "checksum case" ?

two different things, I can make two commits if you prefer.

PR #1582, opened 2022-04-26
new type XC_Icon

gereic:
Added Icons for displaying best result XC-Type as icon in status-bar and on info-page


brunotl (changes requested):
2 file missing in "android-studio/app/src/main/assets/bitmaps/"

brunotl (changes requested):
3 another point : 

- `BmpValue` parameter must be set when `LKFomatValue(...)` is called with index returned by `GetInfoboxIndex(...)`

- `LKFomatValue(...)` is never called with `BmpTitle` parameter

- convert `TType` to `DrawBmp_t` could be moved to a function to avoid duplicated code...

gereic:
In LKDrawLook8000.cpp LKFormatValue is called with bmpTitle.
` Line 570               LKFormatValue(getCustomOverlay(Overlay_RightMid), true, BufferValue, BufferUnit, BufferTitle,&bmpValue,&bmpTitle);`
It is needed, because on Screen when you select best XC-Distance or best XC-Points, it draws the type of XC instead of the Buffertitle.
But only if type is not invalid (CContestMgr::TYPE_INVALID)
On this picture you can see.
At the moment it is only working on right-site, because on left site, the title is not on right of value.

xc_type

gereic:
- want to fix tab-stops and spaces --> revert the changes
- I think now the picture is more readable than a small one, that is the reason for the big size. Otherwise you can't see it maybe.

brunotl:
you can't mix code indent  and code flow change in same commit, otherwise code review become impossible...

still missing :
> BmpValue parameter must be set when LKFomatValue(...) is called with index returned by GetInfoboxIndex(...)

gereic:
Sorry didn't know that.
Changes are made, hopefully all right.

PR #1594, opened 2022-09-06
loading OGN Database after FlarmNet

AlphaLima:
Since OGNs database is widely used (over 20k registrations), it became more important over the FlarmNet database.
This commit also loads the OGN databaste IDs on top of the FlarmNet IDs.
The OGN Data is derived from:
http://ddb.glidernet.org/
the database is not encrypted an can be loaded via
http://ddb.glidernet.org/download/
the OGN database file must be named DATA.OGN and must be placed in the LK8000\_Configuration folder similar to the DATA.FLN file of Flarmnet.

At the moment (Sept. 2022) we have 12365 FlarmNet ID and additional 16807 IDs in OGN. This can be Monitored in the runtime.log.

[078749787] . FLARMNET database, found 12365 IDs
[078749802] . found 8893 IDs also in OGN database
[078749802] . OGN database, found additinal 16807 IDs
[078749802] . total Flarm devices found 29172 IDs

regards
 AlphaLima

### DIFF
--- a/Common/Header/FlarmIdFile.h
+++ b/Common/Header/FlarmIdFile.h
@@ -16,6 +16,7 @@ constexpr size_t FLARMID_SIZE_FREQ = 8;
 
 struct FlarmId {
   explicit FlarmId(const std::string& string);
+  explicit FlarmId(void);
 
   TCHAR id[FLARMID_SIZE_ID] = _T("");
   TCHAR name[FLARMID_SIZE_NAME] = _T("");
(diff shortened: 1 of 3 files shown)


brunotl:
why don't juste use this :  https://github.com/glidernet/ogn-ddb#downloaddownload-flnphp

AlphaLima:
> why don't juste use this : https://github.com/glidernet/ogn-ddb#downloaddownload-flnphp

This Tool didn't work properly for me and we can't use both databases at the same time, using this converter. (There are still some in FlarmNet but not on OGN.)
The FlarmNet Database has a bit more information (Name & Frequency). So the OGN is used for missing in FlarmNet only in my version. There are also some plausibility checks needed for the OGN db since it has some empty entries.

brunotl (commented):
...

PR #1601, opened 2022-10-08
add USB VID/PID pairs for SoftRF Balkan and Prime Mk3 Editions into 'white list' of USB devices

lyusupov:


lyusupov:
Thank you, Bruno!

PR #1610, opened 2023-03-17
Fixes BLUETOOTH external devices on some manufacturers. (Poco, Xiaomi...

andrebstv:
Some manufactures incorrectly implement their permissions, so this is added in order to enable external devices to work on them.

### DIFF
--- a/android/AndroidManifest.xml
+++ b/android/AndroidManifest.xml
@@ -3,7 +3,7 @@
           xmlns:tools="http://schemas.android.com/tools"
           android:installLocation="auto">
 
-  <!-- Required even on later versions (SDK>33) for some manufactures (Xiaomi) -->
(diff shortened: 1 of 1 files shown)


brunotl:
glm and doctest are submodule, there source code won't be commited ...

brunotl:
include wrong glm and doctest submodule upgrade or downgrade ?

2 possibility : 
 - it's upgrade -> make commit to for each submodule upgrade and only for change related to it.
 - it's downgrade -> same a previous and explain why downgrade ...

in both case, that donc look related to bluetooth permission so, can't be include in this pull request ...

brunotl:
manually merged to remove 'glm' and 'doctest' library downgrade

brunotl:

PR #1617, opened 2023-05-11
Update device.cpp, ignore executable rights for native serial RS232

iltis42:
In order to see all serial devices on android devices that have more than one serial port, this change fixes an epic bug that required executable rights on device descriptors that makes no sense for a character device and is not present in current android OS.


iltis42:
The fix solved this issue in XCSoar since a while, making all tty devices accessible:
https://github.com/XCSoar/XCSoar/commit/e042bb63a32a810d6ff1262565fba4035ef95daa

brunotl:
thanks

PR #1641, opened 2024-09-16
New InfoBox parameter - Average Ground Speed.

RyszardLewandowski:
New InfoBox param - Average Ground Speed calculated in real time for the last seconds of the flight determined by the Average Efficiency period.


brunotl:

PR #1644, opened 2024-09-20
New InfoBoxes: NextAverageETE, NextAverageETA.

RyszardLewandowski:
New InfoBoxes: NextAverageETE, NextAverageETA - calculated on an average GS.

### DIFF
--- a/Common/Source/Calc/WaypointArrivalAltitude.cpp
+++ b/Common/Source/Calc/WaypointArrivalAltitude.cpp
@@ -18,6 +18,11 @@ void simpleETE(NMEA_INFO *Basic, DERIVED_INFO *Calculated, int i) {
    if (!ValidWayPoint(i)) return;
 
    WayPointCalc[i].NextETE= WayPointCalc[i].Distance / Basic->Speed;
+   if (Calculated->AverageGS>0) {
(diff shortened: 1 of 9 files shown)


Common/Source/Calc/WaypointArrivalAltitude.cpp
```diff
@@ -18,6 +18,11 @@ void simpleETE(NMEA_INFO *Basic, DERIVED_INFO *Calculated, int i) {
    if (!ValidWayPoint(i)) return;
 
    WayPointCalc[i].NextETE= WayPointCalc[i].Distance / Basic->Speed;
+   if (Calculated->AverageGS>0) {
+       WayPointCalc[i].NextAvrETE = WayPointCalc[i].Distance / Calculated->AverageGS;
+   } else {
+       WayPointCalc[i].NextAvrETE= WayPointCalc[i].NextETE;
```
brunotl, line 24: could be better to set negative value, like that, you draw "---" in the info box if no Avg ground speed is available...

brunotl: "---" until no data available is better than a value with other meaning....

RyszardLewandowski: Right, changed.

PR #1645, opened 2024-11-10
Fix dockerfile

fredgarr:
Hello,
It seems dockerfile has not been updated along with the KOBO rootfs build commit.

Best regards,
Frederic


brunotl:
thanks

PR #1664, opened 2025-10-29
Multi-Leg Pirker analyser

Kamaltdinov:
- Fixed MC estimation during final glide
- Fixed transition to final glide
- Fixed arrival altitude estimation considering previous turn points

### DIFF
--- a/Common/Source/Calc/Pirker.cpp
+++ b/Common/Source/Calc/Pirker.cpp
@@ -61,16 +61,14 @@ double ComputeTaskRequiredAltitude(NMEA_INFO* Basic, DERIVED_INFO* Calculated,
 
     TaskAltitudeRequired += LegAltitude;
 
-    if (true || ISPARAGLIDER) {
-      // if required altitude is less than previous turpoint altitude,
-      //   use previous turn point altitude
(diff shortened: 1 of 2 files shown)


Kamaltdinov:
> what is your compiler version ?

(Debian 8.3.0-6) 8.3.0

Common/Source/Calc/Pirker.cpp
```diff
@@ -8,16 +8,105 @@
...
+
+    double this_LegTimeToGo;
+    double LegAltitude = GlidePolar::MacCreadyAltitude(
+        this_maccready, NextLegDistance, NextLegBearing, Calculated->WindSpeed,
+        Calculated->WindBearing, 0, 0, this_is_final, &this_LegTimeToGo,
+        height_above_finish, CRUISE_EFFICIENCY);
+
+    TaskAltitudeRequired += LegAltitude;
+
+    if (true || ISPARAGLIDER) {
```
brunotl: always true, need to be removed ( or reason to not remove must be esplained in comment )

brunotl (changes requested):

Issue #1678, opened 2026-03-01
Kobo GLO HD – Automatic Zoom Reset to 350m Causes CPU Saturation and UI Freeze


Ziozanna:
**Description**

I have identified a reproducible issue on the Kobo GLO HD that leads to CPU saturation and makes the system temporarily unresponsive.
The problem occurs when the application automatically resets the zoom level to 350 m under specific conditions.

**Steps to Reproduce**

- Start from the Default View with any zoom level set above 1 km.
- Switch to Side Track MM1 (the only side track view enabled).
- Open a target from the Custom Menu.
- Press OK to close the target; the system returns to the MM1 view.
- Switch back to Default View.
- Here zoom are set to 350m and the system is unresponsive

**Observed Behavior**

- The zoom level is automatically reset to 350 m.
- CPU usage spikes significantly.
- The UI becomes unresponsive.
- The system appears frozen.

If the user manually increases the zoom level (in my setup the zoom control is positioned at the bottom of the map) and waits patiently, the system progressively regains responsiveness as the zoom increases:

- 500 m → still frozen
- 750 m → still high degraded
- ~1.5 km → system returns to normal operation

**Expected Behavior**

- The application should preserve the previously selected zoom level when switching views.
- The system should never automatically set the zoom to a level that causes excessive CPU load.
- The UI must remain responsive during view transitions.

**Impact**

The Kobo GLO HD hardware is not powerful enough to handle such deep zoom levels without severe performance degradation.
In a flight scenario, there is no time to diagnose or wait for the system to recover. A temporary UI freeze represents a significant operational risk.

Additional Information

A YouTube video demonstrating the issue is attached. (pls note the clock seconds frozen on 350m zoom)
https://youtube.com/shorts/VQaU2nRFt5I?si=yYKz6IEqM1mggY8p

PR #1683, opened 2026-03-05
fix issue #1678

brunotl:
- CPU load at high zoom level
- Data race on map mode and zoom
- Invalid map rendreing in Target dialog


brunotl:
@coderabbitai full review

PR #1693, opened 2026-04-14
Update Model.cpp add ID for my Kobo Clara HD Model

tvebi:
Set up LK8000 7.5.0 on my new Kobo Klara HD. Wifi did not work because the model was not detected as Clara HD so no driver was loaded.
The model number I found is "SN-R2493" so I added this to Model.cpp. Verified this for version  LK8000-7.4.24


tvebi:
Thank you for adding.

Best Regards 
Eberhard

PR #1696, opened 2026-05-24
Fix CUP parser ignoring '*' comment lines during format detection

### LINKED ISSUE #1695: CUP parser fails to load waypoints when file starts with '*' comment lines
## Description

A valid `.cup` waypoint file that begins with `*` comment lines (e.g. author info, date, format notes) is silently rejected by the parser: **zero waypoints are loaded**.

## Root cause

`ReadWayPointFile()` in `Common/Source/Waypoints/ReadFile.cpp` has two loops:

1. **Format detection loop** — reads lines until it recognises the file format (CUP, OZI, DAT, …).
2. **Data reading loop** — reads and parses waypoint entries; already skips `*` comment lines.

The format detection loop skips empty lines and lines starting with `B` (encoding marker), but **does not skip `*` comment lines**. When the first non-empty line starts with `*`, the loop exits without ever finding the `name,code,country` header, leaving `cup_header` uninitialised. With an empty `cup_header`, every subsequent field lookup returns an empty string, latitude/longitude validation fails, and no waypoints are loaded.

## Example

A file like:
```
* WAYPOINTS ITALY 2026
* Author: ...
*
name,code,country,lat,lon,elev,style,rwdir,rwlen,freq,desc
SOME AIRFIELD,...
```
loads **0 waypoints**, while the same file without the leading comment lines loads correctly.

## Fix

Add the same `*` guard to the format detection loop, mirroring what already exists in the data reading loop:

```cpp
if (src_line.front() == '*') {
  continue;  // Skip comment
}
```

Tested on Linux and Android (debug build).

Ziozanna:
Fixes #1695

## Problem

\`ReadWayPointFile()\` in \`Common/Source/Waypoints/ReadFile.cpp\` has two distinct loops:

1. **Format detection loop** — reads lines until it recognises the file format (\`name,code,country\` header for CUP, \`OziExplorer…\` for OZI, etc.).
2. **Data reading loop** — reads and parses waypoint entries line by line.

The data reading loop correctly skips \`*\` comment lines (line 141):
\`\`\`cpp
if (src_line.front() == '*') {
  continue;  // Skip comment
}
\`\`\`

The format detection loop, however, only skips empty lines and lines starting with \`B\` (encoding marker). It does **not** skip \`*\` comment lines.

When a \`.cup\` file begins with one or more \`*\` comment lines, the format detection loop encounters the first comment before reaching the \`name,code,country\` header. It does not recognise the comment as a known format signature, so it falls through to the extension-based fallback: it sets \`fileformat = LKW_CUP\` but **exits without initialising \`cup_header\`**.

With \`cup_header\` empty, every subsequent field lookup (\`name\`, \`lat\`, \`lon\`, \`elev\`, …) returns an empty string. Latitude and longitude validation fails for every entry, \`ParseCUPWayPointString()\` returns \`false\` on every line, and **zero waypoints are loaded** from an otherwise perfectly valid file — with no error reported to the user.

## Example

A file like:
\`\`\`
* WAYPOINTS ITALY 2026
* Author: ...
*
name,code,country,lat,lon,elev,style,rwdir,rwlen,freq,desc
SOME AIRFIELD,CS11,IT,3945.416N,01626.300E,2m,2,120,850m,130.000,SOME AIRFIELD
\`\`\`
loads **0 waypoints**, while the identical file without the leading comment lines loads correctly.

## Fix

Add the same \`*\` guard to the format detection loop, mirroring what already exists in the data reading loop:

\`\`\`cpp
if (src_line.front() == '*') {
  continue;  // Skip comment
}
\`\`\`

## Testing

Tested on all three supported platforms with a \`.cup\` file containing leading \`*\` comment lines:

- [x] **Linux** — waypoints load correctly
- [x] **Android** — waypoints load correctly (debug build, Samsung S24)
- [x] **Kobo** — waypoints load correctly

Regression: \`.cup\` files without comment lines behave identically to before.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Common/Source/Waypoints/ReadFile.cpp
+++ b/Common/Source/Waypoints/ReadFile.cpp
@@ -36,6 +36,9 @@ int ReadWayPointFile(std::istream& stream, int fileformat) {
     if (src_line.front() == 'B') {
       continue;  // Skip encoding
     }
+    if (src_line.front() == '*') {
+      continue;  // Skip comment
+    }
 
     if (src_line.starts_with("G  WGS 84") || src_line.starts_with("G WGS 84") ||
         src_line.starts_with("\xEF\xBB\xBFG  WGS 84")) {


brunotl:
no comment line in "cup" files format : https://downloads.naviter.com/docs/SeeYou_CUP_file_format.pdf

what app produce this kind of files ?

Ziozanna:
You are correct that the official CUP specification (Naviter) does not define a comment syntax. This fix is not about adding comment support to the CUP format itself.

The change is in the **format detection** logic (`ReadWayPointFile`), not in the CUP waypoint parser. During format detection, LK8000 scans the first lines of the file to decide whether it is a CUP, SeeYou, or other format. If any of those lines starts with `*`, the detection algorithm currently fails to recognise a perfectly valid CUP file and either misidentifies the format or rejects the file entirely.

In practice, files with `*` lines at the top are produced by:
- Script/Python converters that prepend metadata or attribution lines
- AI-assisted tools that add a generation comment
- Users who manually add a description line following the convention of other formats (OpenAir, IGC) where `*` is a standard comment character

The CUP waypoint data that follows is fully valid and compliant with the spec. The fix simply tells the format detector to skip those lines rather than treating them as format-identifying content, making LK8000 consistent with how other gliding software (XCSoar, etc.) handles real-world files.

No CUP spec behaviour is changed; only the robustness of file ingestion is improved.

PR #1698, opened 2026-05-24
Android: multi-layer GPS fallback for broken HAL on custom ROMs

### LINKED ISSUE #1697: Android: GPS HAL failure on custom ROMs leaves LK8000 without fix
## Summary

On some devices running custom ROMs (e.g. LineageOS on a Samsung Galaxy Tab S2 `gts210velte`), the Qualcomm GPS HAL daemon fails to initialise a power-management sysfs node at startup:

```
LocSvc_eng: E/gps_set_battery_flag() : file open error
```

The HAL is reported as *useable* by the system, but it **never delivers NMEA sentences or location callbacks** to apps via `LocationManager.GPS_PROVIDER`. The result is that LK8000 loops forever in:

> Port A gone silent / NO FIX

even though the GPS chip itself acquires satellites normally (confirmed via a third-party NMEA app that uses its own HAL workaround).

## Root cause

LK8000's `InternalGPS.java` relies exclusively on the `OnNmeaMessageListener` / `addNmeaListener` path. When the HAL is broken those callbacks are never fired, and the `onLocationChanged()` override was empty (a no-op), so there was no fallback.

Google Maps and similar apps are unaffected because they use the Fused Location Provider (Google Play Services), which has its own GPS implementation that bypasses the standard Android HAL.

## Affected devices

Any Android device where the GPS HAL initialises without error but silently discards callbacks. Concretely observed on:

| Device | ROM | Android |
|---|---|---|
| Samsung Galaxy Tab S2 (`gts210velte`) | LineageOS 17.1 | 10 |

## Related issue

Same device also crashed on startup due to OpenMP runtime incompatibility (`__kmp_abort_process`). That was fixed separately in the same PR by adding an `ENABLE_OPENMP` CMake option (default `ON`).

## Fix

See linked PR for the full implementation. In brief:

1. Track real NMEA activity with `lastNmeaTime`; use it to gate all fallback paths (no change on working devices).
2. Subscribe to `NETWORK_PROVIDER` and `PASSIVE_PROVIDER` in parallel so `onLocationChanged()` fires even when `GPS_PROVIDER` is silent.
3. Implement `onLocationChanged()` to synthesise `$GPRMC`/`$GPGGA` sentences (only when no real NMEA in the last 5 s).
4. Add a 1-second `locationPoller` that calls `getLastKnownLocation()` across all providers to ensure data flow when stationary.
5. Filter `onProviderDisabled/Enabled` to `GPS_PROVIDER` only to prevent spurious disconnects from fallback providers.
6. Terminate synthetic NMEA sentences with `\r\n` to match the real format and display correctly in the serial terminal.

## Testing

- **Samsung Galaxy S24 (Android 15, working HAL)** — no regression, real NMEA path used exclusively
- **Samsung Galaxy Tab S2 `gts210velte` (LineageOS 17.1 / Android 10, broken HAL)** — GPS fix acquired both with WiFi and without WiFi (outdoors)

Ziozanna:
Fixes #1697

## Problem

On some devices running custom ROMs (e.g. LineageOS on Samsung Galaxy Tab S2 `gts210velte`), the Qualcomm GPS HAL daemon fails silently at startup:

```
LocSvc_eng: E/gps_set_battery_flag() : file open error
```

The HAL is marked *useable* by the system but **never delivers NMEA sentences or `onLocationChanged()` callbacks** to apps via `LocationManager.GPS_PROVIDER`. LK8000 loops forever in "Port A gone silent / NO FIX".

The same device also crashed immediately on startup due to OpenMP runtime incompatibility (`__kmp_abort_process` inside `libomp.so`).

## Root cause

`InternalGPS.java` relied exclusively on `OnNmeaMessageListener`. When the HAL is broken those callbacks never fire, and `onLocationChanged()` was an empty override — no fallback existed.

The OpenMP crash was caused by the `libomp.so` bundled in the NDK being incompatible with the device's custom ROM kernel.

### `android/src/org/LK8000/InternalGPS.java`

1. **`lastNmeaTime` tracking** — volatile timestamp set on every real NMEA sentence. All fallback paths are gated behind `System.currentTimeMillis() - lastNmeaTime > NMEA_TIMEOUT_MS` (5 s). On devices with working HALs this is always false, so the fallback code is dead.

2. **Fallback provider subscriptions** — when `GPS_PROVIDER` is selected, also subscribe to `NETWORK_PROVIDER` and `PASSIVE_PROVIDER`. PASSIVE piggybacks on other apps' location requests (Google Maps, Enroute) at zero extra battery cost. All three subscriptions are cleaned up together via `removeUpdates(this)`.

3. **`onLocationChanged()` implemented** — synthesises `$GPRMC` + `$GPGGA` sentences from the `Location` object, but only when no real NMEA has been received in the last 5 s.

4. **`locationPoller` Runnable** — runs every 1 s; calls `getLastKnownLocation()` across all providers and synthesises NMEA from the freshest fix (max age 60 s). Ensures continuous data flow when the device is stationary and providers stop pushing updates.

5. **`onProviderDisabled/Enabled` filtered** — now only reacts to `GPS_PROVIDER`. Previously any provider change (e.g. NETWORK going offline) would call `setConnectedSafe(false)` on working devices.

6. **Synthetic NMEA format** — sentences are terminated with `\r\n` to match real NMEA and display correctly in the LK8000 serial terminal.

### `CMakeLists.txt` + `android-studio/app/build.gradle`

Add `ENABLE_OPENMP` CMake option (default `ON`). Builds for devices where the OpenMP runtime crashes can be produced with `-PENABLE_OPENMP=OFF` (or `./gradlew app:assembleDebug -PENABLE_OPENMP=OFF`). No change to builds that do not set the flag.

## Impact on other Android devices

| Scenario | Behaviour |
|---|---|
| Working GPS HAL (e.g. S24, Pixel) | `lastNmeaTime` is updated every ~1 s; all fallback paths are never taken. No regression. |
| HAL broken, WiFi/network available | `onLocationChanged()` + poller synthesise NMEA from network location. |
| HAL broken, WiFi off, outdoors | Poller picks up `getLastKnownLocation(GPS)` once the HAL eventually delivers a cached fix; other apps (e.g. maps) keep PASSIVE provider alive. |
| Stationary, no other apps | Poller covers the window after providers stop pushing updates. |

## Testing

- **Samsung Galaxy S24** (Android 15, stock ROM, working HAL) — no regression; real NMEA path used exclusively.
- **Samsung Galaxy Tab S2 `gts210velte`** (LineageOS 17.1 / Android 10, broken HAL) — GPS fix acquired outdoors both with WiFi on and with WiFi off.
- NMEA terminal in LK8000 shows correctly formatted sentences with line breaks between fixes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,11 +35,13 @@ if("${ANDROID_ABI}" MATCHES "arm")
     )
 endif()
 
+option(ENABLE_OPENMP "Enable OpenMP support" ON)
+
 add_compile_options(
     -Werror
     -Wall
     -fno-math-errno
-    -fopenmp
+    $<$<BOOL:${ENABLE_OPENMP}>:-fopenmp>
     -fsigned-char
     -Wunused-label
     -Wunused-variable
@@ -943,6 +945,6 @@ target_link_libraries(LK8000 PRIVATE
     z
     EGL
     GLESv2
-    -fopenmp
-    -static-openmp
+    $<$<BOOL:${ENABLE_OPENMP}>:-fopenmp>
+    $<$<BOOL:${ENABLE_OPENMP}>:-static-openmp>
 )
--- a/Common/Source/Dialogs/dlgProgress.cpp
+++ b/Common/Source/Dialogs/dlgProgress.cpp
@@ -115,8 +115,10 @@ void dlgProgress::OnProgressPaint(WindowControl * Sender, LKSurface& Surface) {
 
   InflateRect(&PrintAreaR, -NIBLSCALE(2), -NIBLSCALE(2));
 
-  const TCHAR* text = Sender->GetCaption();
-  Surface.DrawText(text, &PrintAreaR, DT_VCENTER|DT_SINGLELINE);
+  if (PrintAreaR.bottom > PrintAreaR.top && PrintAreaR.right > PrintAreaR.left) {
+    const TCHAR* text = Sender->GetCaption();
+    Surface.DrawText(text, &PrintAreaR, DT_VCENTER|DT_SINGLELINE);
+  }
 
   Surface.SelectObject(ohB);
   Surface.SelectObject(ohP);
--- a/Common/Source/Waypoints/ReadFile.cpp
+++ b/Common/Source/Waypoints/ReadFile.cpp
@@ -36,6 +36,9 @@ int ReadWayPointFile(std::istream& stream, int fileformat) {
     if (src_line.front() == 'B') {
       continue;  // Skip encoding
     }
+    if (src_line.front() == '*') {
+      continue;  // Skip comment
+    }
 
     if (src_line.starts_with("G  WGS 84") || src_line.starts_with("G WGS 84") ||
         src_line.starts_with("\xEF\xBB\xBFG  WGS 84")) {
--- a/Common/Source/xcs/Screen/OpenGL/SubCanvas.cpp
+++ b/Common/Source/xcs/Screen/OpenGL/SubCanvas.cpp
@@ -43,8 +43,8 @@ SubCanvas::SubCanvas(Canvas &canvas, RasterPoint _offset, PixelSize _size)
   offset = canvas.offset + _offset;
   
   /* sub canvas bottom right limits can't be outside "parent" canvas. */
-  size.cx = std::min<PixelScalar>(_size.cx, canvas.GetSize().cx - _offset.x) ;
-  size.cy = std::min<PixelScalar>(_size.cy, canvas.GetSize().cy - _offset.y) ;
+  size.cx = std::max<PixelScalar>(0, std::min<PixelScalar>(_size.cx, canvas.GetSize().cx - _offset.x));
+  size.cy = std::max<PixelScalar>(0, std::min<PixelScalar>(_size.cy, canvas.GetSize().cy - _offset.y));
 
   if (relative.x != 0 || relative.y != 0) {
     OpenGL::translate += _offset;
@@ -61,8 +61,9 @@ SubCanvas::SubCanvas(Canvas &canvas, RasterPoint _offset, PixelSize _size)
 #endif /* !USE_GLSL */
   }
   
-  if ((relative.x + size.cx < canvas.GetSize().cx) 
-          || (relative.y + size.cy < canvas.GetSize().cy)) {
+  if (size.cx > 0 && size.cy > 0 &&
+      ((relative.x + size.cx < canvas.GetSize().cx)
+          || (relative.y + size.cy < canvas.GetSize().cy))) {
 
     /* Enable Clipping */
     push_scissor = std::make_unique<GLPushScissor>();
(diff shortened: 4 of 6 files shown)


Ziozanna:
## Additional fix: OpenGL SubCanvas crash on high-DPI Android screens

While testing the Direct-To autopilot feature on a **Samsung Galaxy Tab** (2048×1536, 320 dpi, LineageOS 17.1), a second crash was found that affects any Android device with a high-DPI display — independent of the GPS HAL issue fixed by this PR.

### Root cause

`SubCanvas::SubCanvas` clamps its size to the parent canvas bounds using `std::min<PixelScalar>`. When the SubCanvas offset exceeds the parent canvas dimensions (which can happen with high-DPI dialog controls), the result of `canvas.size - offset` is negative. `std::min` of a positive requested size and a negative available space returns the negative value, which is then passed as `GLsizei height` to `GLScissor` → `assert(height > 0)` → abort.

The same assert fires in `dlgProgress::OnProgressPaint` when two consecutive `InflateRect(-NIBLSCALE(2), -NIBLSCALE(2))` calls reduce the text rect height to zero or below on high-DPI screens.

### Fix (added to this PR in commit `d0ad5c8`)

**`Common/Source/xcs/Screen/OpenGL/SubCanvas.cpp`** — clamp size to a minimum of 0; skip GLCanvasScissor creation when size is zero (control is fully outside the visible canvas):

```cpp
size.cx = std::max<PixelScalar>(0, std::min<PixelScalar>(_size.cx, canvas.GetSize().cx - _offset.x));
size.cy = std::max<PixelScalar>(0, std::min<PixelScalar>(_size.cy, canvas.GetSize().cy - _offset.y));

if (size.cx > 0 && size.cy > 0 && (...)) {
    push_scissor = std::make_unique<GLPushScissor>();
    scissor     = std::make_unique<GLCanvasScissor>(*this);
}
```

**`Common/Source/Dialogs/dlgProgress.cpp`** — guard `DrawText` against a degenerate rect:

```cpp
if (PrintAreaR.bottom > PrintAreaR.top && PrintAreaR.right > PrintAreaR.left) {
    Surface.DrawText(text, &PrintAreaR, DT_VCENTER|DT_SINGLELINE);
}
```

### Why it belongs here

Without this fix, any device where this PR resolves the OpenMP startup crash will immediately hit the SubCanvas crash when the first dialog or map overlay is painted. The two fixes are independent of the GPS HAL fallback logic and have no regression risk on working devices.

Ziozanna:
Closing without merging. Investigation showed the root cause on the affected device (Samsung Galaxy Tab S2 gts210velte, LineageOS 17.1) is a missing GSS kernel driver: the GPS HAL requires /dev/gss which is never created because the MSM8952 GSS driver is not compiled into the LineageOS kernel. The fallback logic in this PR helps when network location is available (indoors/WiFi) but cannot provide satellite GPS in flight without network coverage, which was the primary use case. The branch is kept locally for reference.

PR #1705, opened 2026-06-08
GA Direct To: off-task fix navigation with autopilot XTE correction

### LINKED ISSUE #1700: GA navigation: Direct To with temporary origin leg and ATC-relevant target info
## Problem

A GA pilot planning a structured route — multiple waypoints, possibly detouring around airspace or restricted zones — frequently needs to abandon intermediate fixes and navigate direct to a later point. This happens when:

- **ATC clears you direct**: *"[Callsign], cleared direct [fix], report estimating."*
- **The pilot elects to shortcut**: traffic, weather or changing conditions make skipping waypoints the better choice.

LK8000 has no way to express this intent to the navigation engine. Selecting a later waypoint via the Task point selector does not create a new nav reference from the current position — the autopilot XTE output continues to reference the original task leg, which is now irrelevant. The aircraft is steered toward a ghost track instead of the new direct destination.

## Expected behaviour

1. The pilot selects any task waypoint ahead of the active one.
2. A **Direct To** action captures the **current aircraft position** as the origin of a temporary leg toward the selected waypoint.
3. The autopilot $GPRMB output immediately reflects XTE relative to this new direct leg.
4. The temporary leg is silently discarded when the task advances normally or is reset.

A brief confirmation step between pressing Direct To and committing the change would be desirable to prevent accidental activation.

## Additional need — ATC reporting fields in the Target dialog

When navigating direct to a new fix, the pilot must report to ATC. The current Target dialog provides no useful information for this.

For GA aircraft the dialog should show:
- **Distance** to the selected waypoint (direct, from current position)
- **AvETE** — estimated time enroute based on average ground speed
- **AvETA** — estimated local arrival time

These are the values needed to answer *"report estimating"* without mental arithmetic.

Ziozanna:
Closes #1700

## Overview

Adds a **Direct To** function exclusively for GA fixed-wing aircraft (`ISGAAIRCRAFT`). The pilot selects any waypoint during task flight and activates Direct To; the aircraft navigates to that fix, then automatically resumes the task from the nearest task point. All glider code paths are completely unchanged.

## Why GA-only

GA aircraft use autopilots that track **XTE (cross-track error)** relative to a defined leg (origin → destination). The autopilot needs a fixed reference leg to hold course; this is meaningless for gliders navigating thermalling tasks. Separating GA logic into dedicated files (`GADirectTo.cpp`, `dlgTargetGA.cpp`) keeps the glider code clean and makes the GA extension easy to review and maintain.

## User flow

1. During task flight, pilot opens the waypoint quick dialog (map tap or pan view) on any fix
2. **"Direct To"** button appears (GA mode only; "Goto" on gliders is unchanged)
3. A **10-second countdown dialog** appears showing BRG / DIST / ETE / ETA to the fix — pilot can cancel at any time
4. Navigation switches to the off-task fix; the XTE reference leg is `Origin → Fix`
5. The task is fully preserved and visible on the map
6. On arrival at the fix, navigation resumes from the nearest task point; `DirectToActive` is cleared

## New files

| File | Purpose |
|---|---|
| `Common/Header/GADirectTo.h` | Public API for GA Direct To state |
| `Common/Source/Calc/Task/GADirectTo.cpp` | State management: `DirectToActive`, `DirectToWaypointIndex`, `DirectToOriginLat/Lon`; browse-WP tracking for Target dialog; `GA_GetDirectToNavIndex()`, `GA_GetTargetPanLoopStart()` |
| `Common/Source/Dialogs/dlgTargetGA.cpp` | GA-specific Target dialog tab extracted from `dlgTarget.cpp`: Dist/AvETE/AvETA fields, Prev/Next/Direct To navigation buttons, off-task fix info panel |
| `Common/Source/Dialogs/dlgDirectToCountdown.cpp` | 10-second countdown dialog with live BRG/DIST/ETE/ETA display and Cancel button |

## Modified files — navigation

- **`devGenericAutopilot.cpp`**: XTE leg uses `DirectToOrigin → DirectToFix` when `DirectToActive`, providing the autopilot a clean straight-line reference
- **`DistanceToNext.cpp`, `DoCalculations.cpp`**: navigation calculations target the off-task fix when `DirectToActive`
- **`LKProcess.cpp`, `MapWindow_Utils.cpp`, `OverTargets.cpp`**: display and overtarget index target the off-task fix
- **`InSector.cpp`**: arrival at the off-task fix triggers task resume from the nearest task point and clears `DirectToActive`
- **`AnnounceWPSwitch.cpp`, `ClearTask.cpp`**: clear `DirectToActive` on task transitions to prevent stale state

## Modified files — UI

- **`dlgTarget.cpp`**: GA-specific sections extracted into `dlgTargetGA.cpp`; glider layout unchanged
- **`dlgWayQuick.cpp`**: "Goto" button renamed "Direct To" in GA mode; calls `ShowDirectToOffTaskDialog()`
- **`CTaskFileHelper.cpp`, `LoadGpxTask.cpp`**: use airfield matching for all GA task waypoints so intermediate airports resolve correctly
- **`ExpandMacros.cpp`, `InputEvents.cpp`, `DEFAULT_MENU.TXT`**: wire Direct To to the input event system and menu

## Modified files — map drawing

- **`DrawTask.cpp`**: GA aircraft always renders the full task route from WP[0] (pilot may need to return to any earlier waypoint)
- **`DrawBearing.cpp`** (TARGET_PAN mode): when `DirectToActive`, the bearing chain starts at `ActiveTaskPoint` instead of `ActiveTaskPoint+1`, making the re-join leg (off-task fix → `Task[ActiveTaskPoint]`) visible when the pilot opens the Target dialog on the off-task fix

## i18n

New tokens for "Direct To", countdown labels, Dist/AvETE/AvETA fields. All 24 language files updated (English only; other languages retain existing strings until translators update).

## Testing

Tested on Linux simulator and Android (Samsung Galaxy Tab S2, LineageOS 17.1):
- On-task Direct To (task waypoint): countdown fires, XTE reference updated, task continues normally
- Off-task Direct To (arbitrary fix): navigation switches, task preserved on map, resume on arrival
- Cancel during countdown: navigation unchanged
- Task cleared while DirectTo active: `DirectToActive` cleared correctly

## Docs

User manual attached (Italian + English, v4).


Ziozanna:
**Update 2026-06-11 — PR ready for flight test**

Since the initial submission, the following issues were identified during ground testing and resolved:

**Android crash fixes**
- UTF-8 sanitization of Latin-1 waypoint names in pan Direct To description (signal 6 / `ValidateUTF8` assert)
- Fixed invalid UTF-8 byte `\xb0` (Latin-1 degree) → `\xc2\xb0` in countdown BRG display
- Fixed null `MsgToken<2521>` ("Direct To") crash caused by outdated language files on device

**Target view — Approach button logic**
- Fixed spurious Approach button appearing when Direct To targets an arbitrary map point: the button was falling through to the task waypoint at `ActiveTaskPoint` instead of checking the actual DirectTo target
- Pan Direct To now snaps to the nearest landable waypoint (airport/outlanding) within ~4% of the visible screen range (~30 px zoom-independent); if the pan cursor is near an airport, the real WP index is used so Target view can correctly offer the Approach dialog

**Map rendering in Target pan mode**
- The rejoin leg (DirectTo WP → Task[ActiveTaskPoint]) is now drawn with `DrawMulticolorDashLine` in task color, visually distinguishing it from the solid magenta bearing line and making the return route to task immediately readable

**Status:** ground-tested on Linux desktop and Kobo. Missing: flight test with live GPS and autopilot XTE output.

Ziozanna:
**Fix applied** — commit d1f344c

Applied the suggested fix: replaced `\xc2\xb0` with the `°` character directly in the `_T()` string literal in `dlgDirectToCountdown.cpp`. You were right that the raw byte escape breaks on Windows since `_T()` widens to `L"..."` and `\xc2\xb0` gets treated as U+C2B0 instead of the degree symbol U+00B0.

Tested on:
- **Linux** (`TARGET=LINUX`): builds clean, no warnings
- **Windows 64-bit** (`TARGET=PCX64`, MinGW 14-posix cross-compile + Wine): builds clean, runs correctly under Wine — degree symbol renders correctly in the countdown BRG display

Ziozanna:
**Flight test — 2026-06-13 ✓**

Tested in real flight on Kobo Glo HD with BlueFlyVario connected to Flybox Oblo2 autopilot:

- Called Direct To on an arbitrary map point mid-task
- Confirmed rejoin leg from the arbitrary point back to the next task waypoint
- Navigated back to the previous task point
- Navigated forward through subsequent task points
- Supervised the approach to the Direct To target: the bearing line transitions to a softer leg angle compared to the leg that would originate from the activation point

All scenarios work as designed. Ready for review.

Ziozanna:
**CodeRabbit review — fixes applied**

Addressed 4 of the 13 findings. The remaining items were evaluated and either intentional by design or not applicable:

- **Translations (`_@M002521_`):** "Direct To" is kept in English across all locales — it is standard aviation terminology used universally regardless of cockpit language.
- **`dlgTargetGA.cpp` Prev button lower bound:** intentionally allows navigating to previous task points. GA off-task navigation is not race/glider-constrained; a pilot may legitimately return to an earlier waypoint.
- **`dlgTarget.cpp` target_point clamp for GA:** same reason — GA pilots can browse backwards.

**Fixes applied (commit 87a3c55):**

- `CMakeLists.txt`: changed `ENABLE_OPENMP` default from `ON` to `OFF`, consistent with Makefile policy and avoiding unconditional `-static-openmp` linkage on Android and non-Linux toolchains.

- `GADirectTo.cpp`: on off-task arrival (1000 m radius), now also sets `DirectToActive = false` in addition to resetting `DirectToWaypointIndex = -1`, keeping the full DirectTo state consistent with all other reset paths.

- `dlgDirectToCountdown.cpp`: the cancel/early-exit rollback now saves and restores the complete DirectTo state (`active`, `wp_index`, `origin_lat`, `origin_lon`) instead of only `DirectToWaypointIndex`. Also fixed the no-nearby-waypoint coordinate fallback to compute correct N/S and E/W hemisphere suffixes from coordinate sign rather than always printing "N E".

- `dlgTarget.cpp`: the Approach button now appears while browsing off-task waypoints in the Target dialog. `GA_GetApproachWPForTarget()` returns -1 during browse; the previous `ga_offtask` guard blocked the fallback to the task point entirely. A pilot browsing to a task airport can now select the approach directly from the browse view — selecting Approach clears the task, which is the expected and accepted behaviour.

Common/Source/Draw/DrawBearing.cpp
```diff
@@ -69,16 +74,24 @@ void MapWindow::DrawBearing(LKSurface& Surface, const RECT& rc, const ScreenProj
...
 
     if (UseAATTarget() && ( DoOptimizeRoute() || ((ActiveTaskPoint>0) && ValidTaskPoint(ActiveTaskPoint+1))) ) {
       targetLat = Task[ActiveTaskPoint].AATTargetLat;
-      targetLon = Task[ActiveTaskPoint].AATTargetLon; 
+      targetLon = Task[ActiveTaskPoint].AATTargetLon;
     } else {
-      targetLat = WayPointList[Task[ActiveTaskPoint].Index].Latitude;
-      targetLon = WayPointList[Task[ActiveTaskPoint].Index].Longitude;
+      targetLat = WayPointList[overindex].Latitude;
+      targetLon = WayPointList[overindex].Longitude;
```
brunotl, line 99: using overindex here look like a regression ...

Common/Header/Globals.h
```diff
@@ -449,6 +449,10 @@ GEXTERN int PanTaskEdit;
 GEXTERN int RealActiveWaypoint;
 GEXTERN int ActiveTaskPoint;
 GEXTERN bool TaskAborted;
+GEXTERN bool DirectToActive GEXTFALSE;
+GEXTERN double DirectToOriginLat GEXTZERO;
+GEXTERN double DirectToOriginLon GEXTZERO;
+GEXTERN int DirectToWaypointIndex;  // >=0: off-task fix (GA); -1: use Task[ActiveTaskPoint]
```
brunotl: should be initialized to -1

Common/Header/Globals.h
```diff
@@ -449,6 +449,10 @@ GEXTERN int PanTaskEdit;
 GEXTERN int RealActiveWaypoint;
 GEXTERN int ActiveTaskPoint;
 GEXTERN bool TaskAborted;
+GEXTERN bool DirectToActive GEXTFALSE;
+GEXTERN double DirectToOriginLat GEXTZERO;
+GEXTERN double DirectToOriginLon GEXTZERO;
```
brunotl, line 454: instead of two double,  use one GeoPoint

Common/Source/Draw/MapWindow_Utils.cpp
```diff
@@ -268,6 +268,36 @@ void MapWindow::SetTargetPan(bool do_pan, int target_point, unsigned dlgSize /*
...
+  if (dlgSize)
+    targetPanSize = dlgSize;
+
+  if (!mode.Is(Mode::MODE_TARGET_PAN)) {
+    old_latitude = PanLatitude;
+    old_longitude = PanLongitude;
+  }
+
+  TargetPanIndex = -1;  // not a task point
+
```
brunotl, line 286: locking CritSec_TaskData is required, ( not locked by all callers .. )

Common/Source/ExpandMacros.cpp
```diff
@@ -546,6 +546,12 @@ bool ExpandMacros(const TCHAR *In, TCHAR *OutBuffer, size_t Size){
 				lk::snprintf(OutBuffer,Size, _T("%s"),MsgToken<2389>());  //	"NoDraw\nXC"
 		break;
 
+		case 42: // Pan mode Direct To button — visible only in GA aircraft mode
+			if (ISGAAIRCRAFT)
+				lk::strcpy(OutBuffer, _T("Direct\nTo"), Size);
```
brunotl: use MsgToken<...> instead of string literal

Common/Source/InputEvents.cpp
```diff
@@ -865,6 +866,30 @@ void InputEvents::eventMarkLocation(const TCHAR *misc) {
...
+  short th = WithLock(RasterTerrain::mutex, [&]() {
+    return RasterTerrain::GetTerrainHeight(pan_lat, pan_lon);
+  });
+  if (th == TERRAIN_INVALID) th = 0;
+
+  {
+    const std::lock_guard lock(CritSec_TaskData);
+    WayPointList[RESWP_PANPOS].Latitude  = pan_lat;
+    WayPointList[RESWP_PANPOS].Longitude = pan_lon;
+    WayPointList[RESWP_PANPOS].Altitude  = (double)th;
```
brunotl: useless explicit cast

CMakeLists.txt
```diff
@@ -35,11 +35,13 @@ if("${ANDROID_ABI}" MATCHES "arm")
     )
 endif()
 
+option(ENABLE_OPENMP "Enable OpenMP support" OFF)
```
brunotl: why disabled by default ? 
what device require to disable openmp ?

Common/Source/Dialogs/dlgDirectToCountdown.cpp
```diff
@@ -0,0 +1,298 @@
...
+    TCHAR lat_hemi = pan_lat >= 0 ? _T('N') : _T('S');
+    TCHAR lon_hemi = pan_lon >= 0 ? _T('E') : _T('W');
+    lk::snprintf(buf, bufsz, _T("%.4f %c %.4f %c"),
+                 std::fabs(pan_lat), lat_hemi, std::fabs(pan_lon), lon_hemi);
+    return;
+  }
+
+  // Sanitize: waypoint names from Latin-1 databases contain non-UTF8 bytes
+  for (TCHAR* p = ref->name; *p; ++p)
+    if ((unsigned char)*p > 0x7F) *p = '?';
```
brunotl: why this loop ?

if Waypoint Name contains Latin-1 string there is Bug in Waypoint loading !

Common/Source/Dialogs/dlgDirectToCountdown.cpp
```diff
@@ -0,0 +1,298 @@
...
+      double d = 0.;
+      DistanceBearing(pan_lat, pan_lon,
+                      WayPointList[i].Latitude, WayPointList[i].Longitude,
+                      &d, nullptr);
+      if (d < best_dist) { best_dist = d; best_idx = i; }
+    }
+    if (best_idx >= 0) {
+      LK_tcsncpy(countdown_wp_name, WayPointList[best_idx].Name, NAME_SIZE - 1);
+      for (TCHAR* p = countdown_wp_name; *p; ++p)
+        if ((unsigned char)*p > 0x7F) *p = '?';
```
brunotl: why this loop ?

if Waypoint Name contains Latin-1 string there is Bug in Waypoint loading !

Common/Source/Dialogs/dlgDirectToCountdown.cpp
brunotl: why this ?

if Waypoint name is Latin-1 string, ther is bug in Waypoint loading !

Common/Source/Dialogs/dlgDirectToCountdown.cpp
brunotl: why this ?

if Waypoint name is Latin-1 string, ther is bug in Waypoint loading !

brunotl (changes requested):
i just did a quick review before testing...

Ziozanna:
Thanks for the review. I've addressed all the points:

**Latin-1 / UTF-8 (the three comments on `dlgDirectToCountdown.cpp`)** — you were right, it wasn't a waypoint-loading bug: `ParseCUP.cpp` and the shared loader in `ReadFile.cpp` already correctly convert Latin-1→UTF-8 via `from_unknown_charset`. The actual bug was that the 4 places copying the waypoint name in the countdown dialog used `LK_tcsncpy`, which truncates by byte count without respecting UTF-8 boundaries — if the cut landed mid-character it produced invalid UTF-8 and triggered the assert. Replaced all 4 call sites with `lk::strcpy` (already used elsewhere in the file), which truncates UTF-8-safely, and removed the two sanitize loops — which, incidentally, only covered 2 of the 4 vulnerable sites, not all of them.

**`MapWindow_Utils.cpp` locking** — confirmed: `SetTargetPanWaypoint` called `ValidWayPointFast()` outside the lock, then only locked for the data access — a real race. Fixed to match the idiom already used further down in the same file (check inside the lock).

**`ExpandMacros.cpp` MsgToken** — replaced the string literal with a proper token (`_@M002522_`), added to `en.json` with the size bumped in `LKLanguage.h`.

**`InputEvents.cpp` cast** — removed, it was redundant.

**`CMakeLists.txt` OpenMP** — fair question: I checked and `-PENABLE_OPENMP` isn't wired to anything in `build.gradle`, so the change was disabling OpenMP for everyone without the intended workaround mechanism actually working. Reverted it — out of scope for this PR; the specific device with OpenMP issues remains a separate problem to solve on its own.

**`Globals.h` / `DrawBearing.cpp`** — on closer look, these two aren't bugs:
- `DirectToWaypointIndex` is already initialized to -1 in `Globals_Init()`, following the same convention as `ActiveTaskPoint`/`PanTaskEdit` two lines above — I just extended the comment to make that explicit.
- The `overindex` usage in `DrawBearing.cpp` isn't an accidental carry-over: `GetOvertargetIndex()`'s `OVT_TASK` case (the only one reachable in that branch) returns exactly `Task[ActiveTaskPoint].Index` when DirectTo is inactive, and the off-task target when it's active — that's how the bearing line follows DirectTo without a separate branch. Added a comment documenting the invariant.

Rebuilt and tested: Linux (working), Windows PCX64 (compiled), Kobo (compiled), Android debug (compiled).

Ziozanna:
Follow-up on the pan DirectTo flow, prompted by CodeRabbit's review of the previous fix commit.

**Issue found:** during the 10s countdown, `DirectToWaypointIndex = -1` doesn't actually detach the autopilot — `GetOvertargetIndex()` falls back to `Task[ActiveTaskPoint].Index` when it's -1, so an already-active DirectTo briefly targets the task point with stale XTE origin data, i.e. a real (if brief) course-change command before the pilot confirms anything. Separately, `RESWP_PANPOS` was overwritten with the new pan coordinates *before* the countdown started, so if a previous pan DirectTo was already using that same slot, Cancel had nothing valid left to restore.

**Fix:** the pan candidate is now staged in `RESWP_UNUSED` — already a reserved, initialized, and genuinely unused waypoint slot — instead of `RESWP_PANPOS`. It's only copied into `RESWP_PANPOS` (and `DirectToWaypointIndex` set) once the pilot confirms. An active DirectTo is therefore left completely untouched for the whole countdown, whether the pilot cancels or lets a *different* DirectTo run alongside it. This also meant the `DirectToWaypointIndex = -1` workaround and its save/restore-on-cancel bookkeeping were no longer doing anything useful, so I removed them (net -8 lines, simpler than before).

One question for you: since `RESWP_UNUSED` now has an actual, specific purpose (staging slot for pan DirectTo candidates), would you rather I rename it to something self-documenting (e.g. `RESWP_GA_PAN_CANDIDATE`), or leave it as `RESWP_UNUSED` to keep the diff minimal? Happy to do either.

Rebuilt and smoke-tested on Linux, Windows PCX64, Android debug, and Kobo.

brunotl:
it's better to remane RESWP_UNUSED to RESWP_PANPOS_CANDIDATE for meaning and maintenability.

Ziozanna:
Renamed RESWP_UNUSED to RESWP_PANPOS_CANDIDATE as suggested — same value, no behavior change, just the name and its comment updated in Defines.h, Virtuals.cpp, dlgDirectToCountdown.cpp, InputEvents.cpp and GADirectTo.h.

Rebuilt and tested on Linux, Windows PCX64, Android debug, Kobo.

**Update:** CodeRabbit re-reviewed the diff after the rename and flagged a leftover stale comment — the waypoint still carried the string `"UNUSED VIRTUAL TURNPOINT"`. Fixed (commit 864981d7e), along with two other comments in the same area (Defines.h and Virtuals.cpp) that were still referring to the old name.

Ziozanna:
**Merge conflict resolved** — commit b1ec34421

This branch was showing as conflicting against `master`. Root cause: the `lib/doctest` submodule pointer had regressed to v2.4.12 (`1da23a3e8`) during an earlier "bring advanced features up to date from build/full" sync, while upstream `master` had already moved it forward to v2.5.3 (`2d0a9359a`) — two different submodule commits on the same path, which git can't auto-merge.

Merged `upstream/master` into this branch and resolved by taking upstream's (newer) doctest pointer. Every other file that changed on both sides (`CMakeLists.txt`, `Defines.h`, `LKProcess.cpp`, `Globals.cpp`, `InputEvents.cpp`, `CTaskFileHelper.cpp`, `Makefile`) merged cleanly with no further conflicts.

Rebuilt on Linux after the merge: clean compile, no errors.

Branch is now mergeable.

